PR #8, opened 2016-08-02
Method in server to return Cache statistics

sakshia125:
Created a new method in server which returns Cache statistics (Cache Hit Rate and Miss Rate).

### DIFF
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
@@ -27,7 +27,9 @@ import org.apache.thrift.TException;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


rubix-bookkeeper/src/main/thrift/bookkeeper.thrift
```diff
@@ -1,10 +1,14 @@
 namespace java com.qubole.rubix.bookkeeper
+namespace python com.qubole.rubix.bookkeeper
 
 typedef i64 long
+typedef i32 int
```
stagraqubole: is this needed?

sakshia125: no, i forgot to delete it.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -93,6 +104,15 @@ public void setAllCached(String remotePath, long fileLength, long startBlock, lo
         }
     }
 
+    @Override
+    public Map getCacheStats()
+    {
+        Map<String, Double> stats = new HashMap<String, Double>();
+        stats.put("Cache Hit Rate", (double) (cachedRequests/(totalRequests)));
```
stagraqubole: You are loosing precision here. Move the `double` cast inside, ear cachedRequests

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -93,6 +104,15 @@ public void setAllCached(String remotePath, long fileLength, long startBlock, lo
         }
     }
 
+    @Override
+    public Map getCacheStats()
+    {
+        Map<String, Double> stats = new HashMap<String, Double>();
+        stats.put("Cache Hit Rate",  ((double)cachedRequests/totalRequests));
```
stagraqubole: leave space after (double) ... we need to add checkstyle too, will open issue for that

PR #14, opened 2016-08-09
Add checkstyle

sakshia125:
added checkstyle and made corresponding changes

### DIFF
--- a/rubix-hadoop2/src/main/java/com/qubole/rubix/hadoop2/Hadoop2ClusterManager.java
+++ b/rubix-hadoop2/src/main/java/com/qubole/rubix/hadoop2/Hadoop2ClusterManager.java
@@ -43,26 +43,24 @@ import org.apache.hadoop.conf.Configuration;
 public class Hadoop2ClusterManager
         extends ClusterManager
 {
-
     private boolean isMaster = true;
     public int serverPort = 8088;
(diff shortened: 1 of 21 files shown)


rubix-hadoop2/src/main/java/com/qubole/rubix/hadoop2/Hadoop2ClusterManager.java
```diff
@@ -183,24 +179,26 @@ public void setNode(List<Elements> node)
...
+        availMemoryMB    long
+        numContainers    int
         */
 
         String nodeHostName;
         String state;
 
-        String getState() { return state; }
+        String getState()
+        { return state; }
```
stagraqubole: this seems invalid, shouldn't return be on next line and '}' and in line after that?

rubix-hadoop2/src/main/java/com/qubole/rubix/hadoop2/Hadoop2ClusterManager.java
```diff
@@ -183,24 +179,26 @@ public void setNode(List<Elements> node)
...
 
-        String getNodeHostName() { return nodeHostName; }
+        String getNodeHostName()
+        { return nodeHostName; }
```
stagraqubole: same as above

PR #16, opened 2016-08-10
Document hadoop2 plugin usage

sakshia125:


vrajat:
@abhishekdas99 @stagraqubole This seems like useful documentation. Worthwhile to polish and commit ? @abhishekdas99  I have assigned it to you to review and commit.

stagraqubole:
This wasnt merged mainly due to reference to HADOOP_CUSTOM_OVERRIDES which is not used in oss hadoop.

vrajat:
Is this a doc problem only ? Or is there work required to support OSS Hadoop2 ?

stagraqubole:
Just the doc problem, if we provide the right config in core-site.xml, then we dont need to set it via system args.

stagraqubole:
But if the requirement is to use Rubix only for client (say presto) but not for hadoop, then we need to find a way where hadoop conf can be set as part of hadoop jar command

PR #25, opened 2016-09-27
Thrift timeout

sakshia125:
Add timeouts and retries to thrift clients


stagraqubole:
difficult to review with this based over non-local1 and review opened against master.

Is it possible to re-open the PR against non-local1?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
```diff
@@ -0,0 +1,138 @@
...
+                bookKeeperClient.transport.open();
+                return callable.call();
+            }
+            catch (TTransportException e1) {
+                LOG.info("Error while reconnecting");
+                errors++;
+            }
+        }
+
+        if (errors >= maxRetries) {
```
stagraqubole: this if is not needed, this will always be true when we come here

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
```diff
@@ -0,0 +1,138 @@
...
+        super(new TBinaryProtocol(client.transport));
+        this.bookKeeperClient = client;
+        this.maxRetries = maxRetries;
+    }
+
+    @Override
+    public List<Location> getCacheStatus(final String remotePath, final long fileLength, final long lastModified, final long startBlock, final long endBlock, final int clusterType)
+    {
+        try {
+            return bookKeeperClient.getCacheStatus(remotePath, fileLength, lastModified, startBlock, endBlock, clusterType);
```
stagraqubole: why this call? just do retryConnection call

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public void setAllCached(final String remotePath, final long fileLength, final long lastModified, final long startBlock, final long endBlock)
+    {
+        try {
+            bookKeeperClient.setAllCached(remotePath, fileLength, lastModified, startBlock, endBlock);
```
stagraqubole: same like above, no need of this call

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -115,9 +118,10 @@ public FSDataInputStream open(Path path, int bufferSize)
 
         return new FSDataInputStream(
                 new BufferedFSInputStream(
-                        new CachingInputStream(inputStream, this, path, this.getConf(), statsMBean,
-                                               clusterManager.getSplitSize(), clusterManager.getClusterType()),
-                                                    CacheConfig.getBlockSize(getConf())));
+                        new CachingInputStream(inputStream, this, path,
+                                this.getConf(), statsMBean, clusterManager.getSplitSize(),
```
stagraqubole: avoid doing changes which are not relevant to the PR, please revert all changes which are not relevant.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
sakshia125: in retryConnection, there is the logic of closing and opening the bookKeeperClient.transport too.

stagraqubole: thats fine but why do we need to do bookKeeperClient.getCacheStatus once before calling retryConnection? We should be fine just calling retryConnection

sakshia125: no, because this in sort of a wraparound around bookKeeperClient. Calls to getCacheStatus will go through this. So why do we need to retry everytime there's a call to getCacheStatus

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    //Assuming that transport from bookKeeperClient is already open
+    private <V> V retryConnection(Callable<V> callable)
+            throws TException
+    {
+        int errors = 0;
+        try {
+            return callable.call();
```
stagraqubole: again code duplication ... write it such that the while loop below handles it all

rubix-core/src/main/java/com/qubole/rubix/core/RemoteReadRequestChain.java
```diff
@@ -29,10 +29,11 @@
 
 /**
  * Created by stagra on 4/1/16.
- *
+ * <p>
```
stagraqubole: whats this change? please revert

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
```diff
@@ -0,0 +1,130 @@
...
+                @Override
+                public List<Location> call()
+                        throws TException
+                {
+                    return bookKeeperClient.getCacheStatus(remotePath, fileLength, lastModified, startBlock, endBlock, clusterType);
+                }
+            });
+        }
+        catch (TException e) {
+            LOG.info("Could not get cache status from server");
```
stagraqubole: I dont think we need the try-catch here, this method can throw TException

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
```diff
@@ -0,0 +1,130 @@
...
+                public Void call()
+                        throws Exception
+                {
+                    bookKeeperClient.setAllCached(remotePath, fileLength, lastModified, startBlock, endBlock);
+                    return null;
+                }
+            });
+        }
+        catch (Exception e1) {
+            LOG.info("Could not update BookKeeper about newly cached blocks: " + Throwables.getStackTraceAsString(e1));
```
stagraqubole: Same here, no try-catch, let the exception bubble up and let inputStream handle it

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+        catch (Exception e) {
+            LOG.info("Retrying connection" + e.getStackTrace().toString());
+        }
+
+        bookKeeperClient.transport.close();
+
+        while (errors < maxRetries) {
+            try {
+                bookKeeperClient.transport.open();
```
stagraqubole: can you also verify that open() on transport that had been closed is fine? Maybe write a simple java program to do this

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperClient.java
```diff
@@ -53,14 +55,16 @@ public void close()
...
     {
         TTransport transport;
-        transport = new TSocket("localhost", CacheConfig.getServerPort(conf));
+        transport = new TSocket("localhost", CacheConfig.getServerPort(conf), CacheConfig.getClientTimeout(conf));
+
         transport.open();
 
         BookKeeperClient client = new BookKeeperClient(transport);
-        return client;
+        RetryingBookkeeperClient retryingBookkeeperClient = new RetryingBookkeeperClient(client, CacheConfig.getMaxRetries(conf));
```
stagraqubole: Forgot to catch this before, but creating the BookKeeperClient seems redundant. We can very well move the logic of BookKeeperClient.getCacheStatus etc into RetryinBookKeeperClient.

Do we need to keep BookKeeperClient around for any other use? If not lets get rid of it and only keep RetryinBookKeeperClient.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
```diff
@@ -0,0 +1,114 @@
...
+        extends BookKeeperService.Client
+        implements Closeable
+{
+    private static final Logger LOG = LoggerFactory.getLogger(RetryingBookkeeperClient.class);
+    private int maxRetries;
+    private BookKeeperClient bookKeeperClient = null;
+
+    /**
+     * List of causes for TTransportException
+     * ALREADY_OPEN
```
stagraqubole: Why are these here?

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -18,10 +18,9 @@
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
-import com.qubole.rubix.bookkeeper.BookKeeperClient;
 import com.qubole.rubix.bookkeeper.Location;
+import com.qubole.rubix.bookkeeper.RetryingBookkeeperClient;
```
stagraqubole: Looking at all the issues we have seen regarding dependency orders, we should use interfaces instead.
So instead of using RetryingBookkeeperClient in this class we should be using BookKeeperService.Client.

That will help us in future too if we extract out thrift interfaces to spi.

rubix-core/src/main/java/com/qubole/rubix/core/RemoteReadRequestChain.java
```diff
@@ -13,7 +13,7 @@
 package com.qubole.rubix.core;
 
 import com.google.common.base.Throwables;
-import com.qubole.rubix.bookkeeper.BookKeeperClient;
+import com.qubole.rubix.bookkeeper.RetryingBookkeeperClient;
```
stagraqubole: Same like above, better to change to BookKeeperService.Client

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
```diff
@@ -0,0 +1,126 @@
...
+        int errors = 0;
+
+        while (errors < maxRetries) {
+            try {
+                if (!transport.isOpen()) {
+                    try {
+                        transport.open();
+                    }
+                    catch (Exception e1) {
+                        LOG.info("Error while reconnecting");
```
stagraqubole: this could end up in infinite loop, let the transport.open's exception get caught outside so that errors are incremented and we can break

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
```diff
@@ -97,21 +97,17 @@ public Void call()
...
+                    transport.open();
                 }
                 return callable.call();
             }
             catch (Exception e) {
                 LOG.info("Error while connecting" + e.getStackTrace().toString());
                 errors++;
             }
-            transport.close();
+            if (transport.isOpen()) {
```
stagraqubole: why close here? we will close on when close() is called on this object

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RetryingBookkeeperClient.java
```diff
@@ -97,21 +97,17 @@ public Void call()
...
             try {
                 if (!transport.isOpen()) {
-                    try {
-                        transport.open();
-                    }
-                    catch (Exception e1) {
-                        LOG.info("Error while reconnecting");
-                        continue;
-                    }
+                    transport.open();
```
stagraqubole: also, isnt this unncessary? The assumption says transport is already open and also when client is created, transport is opened explicitly as per assumption

PR #33, opened 2016-11-15
Transfer test

sakshia125:
Implement inter-node reads with zero copy


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -117,29 +118,29 @@ public BookKeeper(Configuration conf)
...
 
             if (md.isBlockCached(blockNum)) {
-                blocksInfo.add(Location.CACHED);
+                blockLocations.add(new BlockLocation(Location.CACHED, blockSplits.get(split)));
                 cachedRequests++;
             }
             else {
-                if (localSplits.contains(split)) {
-                    blocksInfo.add(Location.LOCAL);
+                if (currentNodeIndex != -1 && blockSplits.get(split).equalsIgnoreCase(nodes.get(currentNodeIndex))) {
```
stagraqubole: we should return `Direct` type if currentNodeIndex is 1

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -296,6 +297,43 @@ public void onRemoval(
                 .build();
     }
 
+    @Override
```
stagraqubole: add a comment about this method telling that the purpose if only to warm the cache and tell how it is used

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -296,6 +297,43 @@ public void onRemoval(
...
+   public boolean readData(String path, long offset, int length, long fileSize, long lastModified, int clusterType)
+           throws TException
+   {
+       int blockSize = CacheConfig.getBlockSize(conf);
+       byte[] buffer = new byte[blockSize];
```
stagraqubole: Take this buffer out as a static variable of this class.
The data written to it is not used at all and we will waste GC cycles because of so many allocations of this buffer. Better to keep one copy and let all use it to dump data. Add a warning at the declaration of this that data in this buffer will always be garbage due to parallel writes and should not be used anytime

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -296,6 +297,43 @@ public void onRemoval(
...
+       FSDataInputStream inputStream = null;
+       long startBlock = offset / blockSize;
+       long endBlock = ((offset + (length - 1)) / CacheConfig.getBlockSize(conf)) + 1;
+       try {
+           int idx = 0;
+           List<BlockLocation> blockLocations = getCacheStatus(path, fileSize, lastModified, startBlock, endBlock, clusterType);
+
+           for (int blockNum = (int) startBlock; blockNum < endBlock; blockNum++, idx++) {
+               int readStart = blockNum * blockSize;
+               if (blockLocations.get(idx).getLocation() == Location.LOCAL) {
```
stagraqubole: change `blockLocations.get(idx).getLocation() == Location.LOCAL` to `blockLocations.get(idx).getLocation() != Location.CACHED`

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -296,6 +297,43 @@ public void onRemoval(
...
+                   if (fs == null) {
+                       fs = new CachingNativeS3FileSystem(bookKeeperFactory, new Path(path), conf);
+                       inputStream = fs.open(new Path(path), blockSize);
+                   }
+                   inputStream.seek(readStart);
```
stagraqubole: There is a potential optimization to speed up warmup. Right now the work of getCacheStatus is repeated, once in this method and once in inputStream.read. We can rather create a RemoteReadRequestChain here directly and call add(), call() and updateCacheStatus() methods here rather than inputStream.read()

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -337,4 +342,19 @@ static void setCacheDataChosenColumns(Configuration c, int chosen)
...
     }
+
+    public static int getBufferSize(Configuration c)
+    {
+        return c.getInt(bufferSizeConf, bufferSize);
+    }
+
+    public static int getLocalServerPort(Configuration conf)
+    {
+        return conf.getInt(localServerPortConf, 6666);
```
stagraqubole: extract out port number as a const variable in the start like others. And i would suggest to use 8898.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalTransferServer.java
```diff
@@ -0,0 +1,191 @@
...
+        {
+            int port = CacheConfig.getLocalServerPort(conf);
+
+            ExecutorService threadPool = Executors.newCachedThreadPool();
+            try {
+                listener = ServerSocketChannel.open();
+                listener.bind(new InetSocketAddress(port));
+                log.info("Listening on port " + port);
+                while (true) {
+                    SocketChannel clientSocket = listener.accept();
```
stagraqubole: What are the service guarantees for this? Do we know how many connections it can accept? Does it tell when it drops a connection? I hope we have retry in the client code of this, will check when i get there

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalTransferServer.java
```diff
@@ -0,0 +1,191 @@
...
+        {
+            localTransferClient = s;
+        }
+
+        public void run()
+        {
+            try {
+                log.debug("Connected to node - " + localTransferClient.getLocalAddress());
+                BookKeeperFactory bookKeeperFactory = new BookKeeperFactory();
+                ByteBuffer dataInfo = ByteBuffer.allocate(CacheConfig.getDataTransferBufferSize(conf));
```
stagraqubole: is 1024 bytes enough? There can be very long filenames since they are absolute paths

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalTransferServer.java
```diff
@@ -0,0 +1,191 @@
...
+                    }
+                }
+                catch (TException e) {
+                    log.info("Could not create BookKeeper client" + Throwables.getStackTraceAsString(e));
+                    localTransferClient.close();
+                }
+
+                String filename = CacheConfig.getLocalPath(remotePath, conf);
+                FileChannel fc = new FileInputStream(filename).getChannel();
+                int maxCount = (10 * 1024 * 1024);
```
stagraqubole: keep this configurable too? I know 10MB was the best performing number but never know that under different circumstances what value would be better, keeping it configurable will also help us to experiment easily

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
sakshia125: currentNodeIndex is usually -1 for master. So master can read from other nodes, that's why I kept it non-local. But keeping it Direct type should be safer.

rubix-core/src/main/java/com/qubole/rubix/core/NonLocalReadRequestChain.java
```diff
@@ -12,24 +12,143 @@
...
+                        nread = transferClient.read(dst);
+                        bytesread += nread;
+                        totalRead += nread;
+                    }
+                    catch (IOException e) {
+                        e.printStackTrace();
+                        return directReadRequest(readRequests.indexOf(readRequest));
+                    }
+                }
+                if (nread == -1) {
```
sakshia125: nread is the dataRead for the current loop. It's -1 when connection is closed abruptly.

stagraqubole: what if the request was larger than 10MB and first 10MB passed while later it failed?

stagraqubole: got it, was reading it wrong

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -117,29 +122,28 @@ public BookKeeper(Configuration conf)
...
-
             if (md.isBlockCached(blockNum)) {
-                blocksInfo.add(Location.CACHED);
+                blockLocations.add(new BlockLocation(Location.CACHED, blockSplits.get(split)));
                 cachedRequests++;
             }
             else {
-                if (localSplits.contains(split)) {
-                    blocksInfo.add(Location.LOCAL);
+                if (blockSplits.get(split).equalsIgnoreCase(nodes.get(currentNodeIndex))) {
```
stagraqubole: Potential improvement, store Node indices too i.e. split to node index map and compare indices here instead of strings (nodenames). Leave a comment here, we can revisit and see what perf gain we get

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -208,14 +212,56 @@ public void setAllCached(String remotePath, long fileLength, long lastModified,
...
         stats.put("Remote Reads", ((double) remoteRequests));
         stats.put("Non-Local Reads", ((double) (totalRequests - cachedRequests - remoteRequests)));
         return stats;
     }
 
+    //This method is to ensure that data required by another node is cached before it is read by that node
+    //using localTransferServer.
+    // If the data is not already cached, remoteReadRequest for that block is sent and data is cached.
+
+    //TODO : buffer is initialized everytime readData is called and it contains garbage value which is not required. Also, getCacheStatus is called twice.
```
stagraqubole: explain a bit more about why we cannot mark it as static. Add enough context so that when a new person sees this it is easy to pick up the work

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -208,14 +212,56 @@ public void setAllCached(String remotePath, long fileLength, long lastModified,
...
+                    inputStream.read(buffer, 0, blockSize);
+                }
+            }
+            if (inputStream != null) {
+                inputStream.close();
+            }
+            return true;
+        }
+        catch (IOException e) {
+            log.info("Could not cache data: " + Throwables.getStackTraceAsString(e));
```
stagraqubole: close inputStream in the catch too.
Better add a finalize method and close stream there for all cases

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -337,4 +343,20 @@ static void setCacheDataChosenColumns(Configuration c, int chosen)
...
+    {
+        return c.getInt(bufferSizeConf, bufferSize);
+    }
+
+    public static int getLocalServerPort(Configuration conf)
+    {
+        return conf.getInt(localServerPortConf, localServerPort);
+    }
+
+    public static int getDataTransferBufferSize(Configuration conf)
```
stagraqubole: This should be rename, getMaxHeaderSize probably

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -337,4 +343,20 @@ static void setCacheDataChosenColumns(Configuration c, int chosen)
     {
         c.setInt(DATA_CACHE_TABLE_COLS_CHOSEN, chosen);
     }
+
+    //bufferSize for reads in LocalTransferServer
+    public static int getBufferSize(Configuration c)
```
stagraqubole: Rename this too, getLocalTransferBufferSize

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -0,0 +1,180 @@
...
+                    nread += fc.transferTo(position + nread, maxCount, localDataTransferClient);
+                    lengthRemaining = readLength - nread;
+                }
+                if (bookKeeperClient != null) {
+                    bookKeeperClient.close();
+                }
+                fc.close();
+            }
+            catch (Exception e) {
+                log.info("Error in Local Data Transfer Server: " + Throwables.getStackTraceAsString(e));
```
stagraqubole: I think there is log.info("Message", Exception) method which is simpler to use

rubix-spi/src/main/java/com/qubole/rubix/spi/DataTransferClientHelper.java
```diff
@@ -0,0 +1,66 @@
...
+            throws IOException
+    {
+        SocketAddress sad = new InetSocketAddress(remoteNodeName, CacheConfig.getLocalServerPort(conf));
+        SocketChannel sc = SocketChannel.open();
+        sc.configureBlocking(true);
+        sc.socket().connect(sad, CacheConfig.getClientTimeout(conf));
+        return sc;
+    }
+
+    /* order is: int : filePathLength, String : filePath, long : offset, int : readLength, long : fileSize, long : lastModified,
```
stagraqubole: Make this a multiline comment, each arg in one line

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -224,42 +227,61 @@ public Map getCacheStats()
...
-            List<BlockLocation> blockLocations = getCacheStatus(path, fileSize, lastModified, startBlock, endBlock, clusterType);
+            List<BlockLocation> blockLocations = getCacheStatus(remotePath, fileSize, lastModified, startBlock, endBlock, clusterType);
 
             for (int blockNum = (int) startBlock; blockNum < endBlock; blockNum++, idx++) {
                 int readStart = blockNum * blockSize;
+                log.debug(" blockLocation is: " + blockLocations.get(idx).getLocation());
                 if (blockLocations.get(idx).getLocation() != Location.CACHED) {
                     if (fs == null) {
-                        fs = new CachingNativeS3FileSystem(bookKeeperFactory, new Path(path), conf);
-                        inputStream = fs.open(new Path(path), blockSize);
```
stagraqubole: This needs a change in README. Need to inform ppl that the hadoop config while starting BookKeeper should be set correctly to return CachingFileSystem class for remote paths.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -0,0 +1,185 @@
...
+        public void run()
+        {
+            try {
+                log.debug("Connected to node - " + localDataTransferClient.getLocalAddress());
+                BookKeeperFactory bookKeeperFactory = new BookKeeperFactory();
+                ByteBuffer dataInfo = ByteBuffer.allocate(CacheConfig.getMaxHeaderSize(conf));
+
+                int read = localDataTransferClient.read(dataInfo);
+                if (read == -1) {
+                    throw new Exception("Could not read data from Non-local node");
```
stagraqubole: Log this as error, since this is on a separate thread, it might get lost. Add enough info like client name or ip

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -208,14 +215,75 @@ public void setAllCached(String remotePath, long fileLength, long lastModified,
...
+                        inputStream = fs.open(path, blockSize);
+                    }
+                    inputStream.seek(readStart);
+                    inputStream.read(buffer, 0, blockSize);
+                }
+            }
+            return true;
+        }
+        catch (Exception e) {
+            log.info("Could not cache data: " + Throwables.getStackTraceAsString(e));
```
stagraqubole: log.warn with enough info to identify the failing request.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -0,0 +1,185 @@
...
+                    nread += fc.transferTo(position + nread, maxCount, localDataTransferClient);
+                    lengthRemaining = readLength - nread;
+                }
+                if (bookKeeperClient != null) {
+                    bookKeeperClient.close();
+                }
+                fc.close();
+            }
+            catch (Exception e) {
+                log.info("Error in Local Data Transfer Server: ", e);
```
stagraqubole: warn or error type, add info about client

stagraqubole:
Log all exceptions, log them as warning/errors appropriately and add enough context so that the call where the exception occurred can be identified.

rubix-core/src/test/java/com/qubole/rubix/core/TestCachingFileSystem.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+
+/**
+ * Created by sakshia on 25/11/16.
+ */
+
+public class TestCachingFileSystem
```
stagraqubole: Rename to `MockCachingFileSystem`. Classes with Test prefix are usually meant to be the ones carrying test cases

rubix-tests/src/test/java/com/qubole/rubix/tests/TestNonLocalReadRequestChain.java
```diff
@@ -0,0 +1,167 @@
...
+        {
+            public void run()
+            {
+                BookKeeperServer.startServer(conf);
+            }
+        };
+        thread.start();
+
+        while (!BookKeeperServer.isServerUp()) {
+            Thread.sleep(200);
```
stagraqubole: We should add LocalDataTransferServer isUp check here

stagraqubole: nvm, i see we are not starting it here

rubix-tests/src/test/java/com/qubole/rubix/tests/TestNonLocalReadRequestChain.java
```diff
@@ -0,0 +1,167 @@
...
+        TestCachingFileSystem fs = new TestCachingFileSystem();
+        fs.initialize(null, conf);
+        nonLocalReadRequestChain = new NonLocalReadRequestChain("localhost", backendFile.length(), backendFile.lastModified(), conf, fs, backendPath.toString(), ClusterType.TEST_CLUSTER_MANAGER.ordinal());
+    }
+
+    @Test
+    private void testRemoteRead()
+            throws Exception
+    {
+        localDataTransferServer.start();
```
stagraqubole: we should rather wait here till it comes up. Have seen intermittent failures with BookKeeper server if we dont wait

stagraqubole: Do one more change to ensure that we are reading from localDataTransferServer and not direct reads here.
Add to NonLocalRRC a check that if STRICT_MODE is set, then it should fail rather than falling back to direct mode

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
sakshia125: this is logged in catch method

PR #34, opened 2016-12-06
Presto compatibility

sakshia125:
Make master compatible with presto


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -162,17 +164,30 @@ private void initializeClusterManager(int clusterType)
...
                         splitSize = 64 * 1024 * 1024;
                     }
+                    else {
+                        if (clusterType == HADOOP2_CLUSTER_MANAGER.ordinal()) {
+                            clusterManager = new Hadoop2ClusterManager();
+                        }
+
+                        else if (clusterType == PRESTO_CLUSTER_MANAGER.ordinal()) {
+                            try {
+                                nodeName = InetAddress.getLocalHost().getHostAddress();
```
stagraqubole: shouldnt need to do this here. We should just return if the nodeName threw exception earlier and assume it to be set already if we reach here

rubix-presto/src/main/java/com/qubole/rubix/presto/PrestoClusterManager.java
```diff
@@ -12,149 +12,170 @@
...
+                            StringBuffer failedResponse = new StringBuffer();
+
+                            if (allNodesResponseCode == HttpURLConnection.HTTP_OK) {
+                                isMaster = true;
+                                BufferedReader in = new BufferedReader(new InputStreamReader(allHttpCon.getInputStream()));
+                                String inputLine = "";
+                                while ((inputLine = in.readLine()) != null) {
+                                    allResponse.append(inputLine);
+                                }
+                                in.close();
```
stagraqubole: we might leak connections here. If BuffereReader throws exception, we would not close allHttpCon. Same for next request too.
Either refactor so that we can add a finally clause or catch exceptions everywhere appropriately and close required objects.

rubix-presto/src/main/java/com/qubole/rubix/presto/PrestoClusterManager.java
```diff
@@ -12,149 +12,170 @@
...
     public boolean isMaster()
     {
         // issue get on nodesSupplier to ensure that isMaster is set correctly
-        nodesSupplier.get();
+        //nodesSupplier.get();
+        try {
+            nodesCache.get("nodeList");
+        }
+        catch (ExecutionException e) {
+            e.printStackTrace();
```
stagraqubole: isnt this wrong? When we hit an exception, we should let the caller know about it because we dont know if isMaster set is correct or not.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
sakshia125: this is local host address while earlier, we are storing canonical hostname.  In Presto, we need to store the host address while in hadoop and test, we need the hostname

stagraqubole: Got it, leave a comment in that case to make it obvious

rubix-presto/src/main/java/com/qubole/rubix/presto/PrestoClusterManager.java
```diff
@@ -165,35 +208,49 @@ public boolean isMaster()
...
     {
-        return nodesSupplier.get();
+        //return nodesSupplier.get();
+        try {
+            return nodesCache.get("nodeList");
+        }
+        catch (ExecutionException e) {
+            e.printStackTrace();
+        }
+        return null;
```
stagraqubole: we allow returning null but we do handle null in callers. There are two callers for this method, add null check in them

rubix-presto/src/main/java/com/qubole/rubix/presto/PrestoClusterManager.java
```diff
@@ -165,35 +208,49 @@ public boolean isMaster()
     @Override
     public List<String> getNodes()
     {
-        return nodesSupplier.get();
+        //return nodesSupplier.get();
+        try {
+            return nodesCache.get("nodeList");
+        }
+        catch (ExecutionException e) {
+            e.printStackTrace();
```
stagraqubole: log exception

Issue #22, opened 2016-09-15
Central stats collection


stagraqubole:
From master node, we should be able to extract stats from all nodes.

For this, we will need the list of all the nodes in BookKeeper

PR #35, opened 2016-12-09
Stats

sakshia125:
Python client to collect stats


stagraqubole:
Add a section in README on how to build this and use this

vrajat:
@stagraqubole is this related to #22 ? Can you provide more info on requirements and design ?

stagraqubole:
No, this isnt related to that. This is creates a simple python thrift client that can be run locally on any machine running bookkeeper and get the cache status for that machine

PR #49, opened 2017-07-19
Lazy loading of FSDataInputStream

abhishekdas99:
Instead of creating a new instance of FSDataInputStream, we are instantiating it only when we needed.


rubix-tests/src/test/java/com/qubole/rubix/tests/TestCachingInputStream.java
```diff
@@ -22,6 +22,7 @@
 import com.qubole.rubix.bookkeeper.BookKeeperServer;
 import com.qubole.rubix.spi.ClusterType;
 
+
```
stagraqubole: extra newline ... checkstyle will fail when we fix it for tests

stagraqubole:
lgtm, minor checkstyle comment.

Did you run the unit tests? Till we add travis support, we will have to do that manually.

abhishekdas99:
I ran unit tests. It didn't complain. But we can remove this extra line

PR #67, opened 2017-08-16
Get master ip from yarn config.

### LINKED ISSUE #66: Get master IP from hadoop yarn config files
Currently we obtain master IP from master.hostname property which is only populated by Qubole. If this is not present, extract from YARN config files.

pvam:
Fixes #66.

### DIFF
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
@@ -18,6 +18,7 @@ import com.qubole.rubix.spi.BookKeeperService;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.yarn.conf.YarnConfiguration;
 import org.apache.thrift.shaded.server.TServer;
 import org.apache.thrift.shaded.server.TThreadPoolServer;
 import org.apache.thrift.shaded.transport.TServerSocket;
@@ -47,6 +48,7 @@ public class BookKeeperServer
     public static void main(String[] args)
     {
         conf = new Configuration();
+        conf.addResource(new YarnConfiguration());
 
         Runnable bookKeeperServer = new Runnable() {
             public void run()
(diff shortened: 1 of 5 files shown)


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -47,6 +48,7 @@ private BookKeeperServer()
     public static void main(String[] args)
     {
         conf = new Configuration();
+        conf.addResource(new YarnConfiguration());
```
vrajat: What will happen if there is no yarn ? I know thats not possible in Qubole or EMR. Will Rubix fail to start up ? Just for my information. I think thats fine as a start.

pvam: If neither master.hostname or yarn property is not present it will default to locahost (which never happens). I had checked with abhishekmodi that both hadoop1 and hadoop2 will always have this property.

rubix-presto/src/main/java/com/qubole/rubix/presto/PrestoClusterManager.java
```diff
@@ -58,15 +58,16 @@
     private Log log = LogFactory.getLog(PrestoClusterManager.class);
 
     public static String serverPortConf = "caching.fs.presto-server-port";
-    public static String serverAddressConf = "master.hostname";
+    public static String quboleServerAddressConf = "master.hostname";
```
abhishekdas99: Can we change the name of the variable .. Lets not have qubole in the name

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
abhishekdas99: Bookkeeper is common for both hadoop1 and hadoop2. Now sure how hadoop1 cluster will work with yarn configuration .

abhishekdas99: Can the respective cluster managers add the resources they need ? In that case, hadoop2 or presto cluster manager can add YarnConfiguration as resource and it is going to work for hadoop1 as well

pvam: Yes, we can do that as well. Should be a safer change.

pvam:
@abhishekdas99 moved config, can you take a look now?

pvam:
Closing this, please look at #69. @abhishekdas99 just need to check YARN related changes in the new PR.

PR #72, opened 2017-08-22
fix hadoop dependency with rubix-presto

pvam:
This fixes Rubix master UT.

### DIFF
--- a/rubix-presto/pom.xml
+++ b/rubix-presto/pom.xml
@@ -21,11 +21,6 @@
         </dependency>
 
         <dependency>
-            <groupId>com.qubole.rubix</groupId>
-            <artifactId>rubix-hadoop2</artifactId>
-        </dependency>
-
-        <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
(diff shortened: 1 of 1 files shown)


rubix-presto/pom.xml
```diff
@@ -21,9 +21,8 @@
         </dependency>
 
         <dependency>
-            <groupId>org.apache.hadoop</groupId>
```
stagraqubole: This will cause issues in runtime with engine.

pvam:
Spoke offline. Dependency with other modules is not the beat way to do this. Not yet sure why UT is failing with H2.6 dependency where the yarn method exists.

pvam:
It should be fixed now. Will merge after approval.

abhishekdas99:
Did we test this in presto cluster ? If everything is fine, @pvam can we update this with the result. Then I can approve this .. Also get a approval from @stagraqubole as its Presto related. Lets get this done so that we can go ahead with the new release of Rubix

pvam:
Opened #73, closing this. With #73, we dont dependency on yarn anymore we assume the property is populated via rubix client config.

PR #73, opened 2017-08-28
EMR: move client config to independent file

pvam:
Whatever properties that are needed by rubix client can now be specified in pre-populated /etc/rubix/config.xml. Like in qubole, we dump req properties in this file during installation. This avoid the need for an explicit dependency on yarn config.

Tested in Qubole and EMR clusters. We can have a new release now.


pvam:
@vrajat @abhishekdas99 please take a look. planning to merge today

PR #77, opened 2017-09-28
Adding support for rubix scheme and populating filesystem counters

abhishekdas99:


rubix-core/pom.xml
```diff
@@ -15,19 +15,15 @@
 
     <properties>
         <main.basedir>${project.parent.basedir}</main.basedir>
+        <dep.hadoop2.version>2.6.0</dep.hadoop2.version>
```
stagraqubole: this is not required, it is provided by root pom

rubix-core/src/main/java/com/qubole/rubix/core/CachedReadRequestChain.java
```diff
@@ -32,25 +33,27 @@
...
         directBuffer = buffer;
+        this.statistics = statistics;
     }
 
     @VisibleForTesting
     public CachedReadRequestChain(String fileToRead)
             throws IOException
     {
-        this(fileToRead, ByteBuffer.allocate(1024));
+        this(fileToRead, ByteBuffer.allocate(1024), null);
```
stagraqubole: simpler to create new Statistics object here to keep rest of the code simple

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -102,13 +107,17 @@ public void initialize(URI uri, Configuration conf) throws IOException
             throw new IOException("Cluster Manager not set");
         }
         super.initialize(uri, conf);
-        fs.initialize(uri, conf);
+        this.uri = URI.create(uri.getScheme() + "://" + uri.getAuthority());
+        this.workingDir = new Path("/user", System.getProperty("user.name")).makeQualified(this);
```
stagraqubole: is it necessary to set workingDir?

rubix-core/src/main/java/com/qubole/rubix/core/CachedReadRequestChain.java
```diff
@@ -100,6 +103,9 @@ public Integer call()
         log.info(String.format("Read %d bytes from cached file", read));
         fileChannel.close();
         raf.close();
+        if (statistics != null) {
+          statistics.incrementBytesRead(read);
```
stagraqubole: Is there a way that we can provide our own Statistics implementation? Basically, we would want to differentiate amount of data read locally (i.e. in CachenRRC) and amount of data read from other nodes (i.e. in NonLocalRRC).

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -206,7 +219,27 @@ public boolean mkdirs(Path path, FsPermission fsPermission)
     public FileStatus getFileStatus(Path path)
             throws IOException
     {
-        return fs.getFileStatus(path);
+        FileStatus originalStatus = fs.getFileStatus(path);
```
stagraqubole: So changing listing's path to rubix scheme was the main change to support rubix scheme!!

abhishekdas99: Yes this one is main along with initializing  the remotefilesystem with appropriate scheme (for CachingNativeS3FileSystem, its s3n) so that when we are reading from the remote filesystem, it gets reflected properly

rubix-core/pom.xml
abhishekdas99: Done

rubix-core/src/main/java/com/qubole/rubix/core/CachedReadRequestChain.java
abhishekdas99: Yes right now we are only populating bytes read counter for local read. Current implementation of NonLocalRead stats gives you both remote read and non local read combined. But once we have the asynchronous cache warm up feature, it will be easy to differentiate between the data read from the non local cache and data we are downloading in the remote node as part of this request.

stagraqubole: Do you know how are these statistics shown by Hive jobs? i.e. does it call FileSystem.printStatistics method? Based on that we can think of workarounds

abhishekdas99: FileSystem has the statistics object. This is part of hadoop file system level counter. Hive just gets a report from AM and prints it

PR #86, opened 2017-11-27
fix: dev: Local cache file offset should take file split size into account

abhishekdas99:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -184,7 +184,8 @@ public void run()
                 FileChannel fc = new FileInputStream(filename).getChannel();
                 int maxCount = CacheConfig.getLocalTransferBufferSize(conf);
                 int lengthRemaining = readLength;
-                long position = offset;
+                long fileOffset = offset % CacheConfig.getSplitSize(conf);
```
stagraqubole: Can you plz share some context.
From the looks of it, it seems wrong to change the position as the client would have asked to read from this position explicitly.

abhishekdas99: Let say when when we want to read last 10 mb of a 100 mb file, we create the local file and seek to the 90mb position and write 10 mb. In theory the size of the cached file should be 10mb but the actual size of the cached file on disk is 100 mb.

Now say for a 500mb file we want to read last 1mb (footer of an orc file). This file would have two splits. Node one is supposed to have first 256 mb (configured split size) of the file and node two is supposed to have the last 244mb of the file. With the logic mentioned above the file size on the disk will be 500mb in node two. To eliminate this problem partially, we can change convert the actual file read offset to take split size into account. We are saying the actual size of a chunk cannot be more than the split size. So when a request comes in for a read (for the first time or from the cache), we convert the actual offset to the cached file offset. For example, in the above mentioned case the offset of the read would be 499mb, we will convert this to 499%256= 243mb. This is happen while writing to the cache file as well as reading from the cache file. 
The benefit of this change is really significant for a large file. 
Without this change for a 1gb file, the file sizes on disk would be  
N1 - 256mb, N2 - 512 mb, N3 - 768mb abd N4 - 1024mb
With this change
N1 - 256mb, N2 - 256mb, N3 - 256mb, N4 - 256mb

So better utilization of disk space.

stagraqubole: There is no wastage of disk space. These are all sparse files. Check the output of `ls -lash`, it will show that for your example the occupied size on disk is not 500MB but only 1MB which has been brought to disk.
Doing this is error prone as any small calculation error will lead to data corruption and is not saving us anything.

PR #91, opened 2018-01-05
new: usr: Added support for NativeAzure file system

abhishekdas99:


abhishekdas99:
Can we review this one ? I have tested it with wasb:// path.

pom.xml
```diff
@@ -48,7 +48,7 @@
         <dep.airlift.version>0.98</dep.airlift.version>
         <commons-httpclient.version>3.0.1</commons-httpclient.version>
         <dep.hadoop1.version>1.2.0</dep.hadoop1.version>
-        <dep.hadoop2.version>2.6.0</dep.hadoop2.version>
+        <dep.hadoop2.version>2.7.0</dep.hadoop2.version>
```
vrajat: Is the upgrade to 2.7.0 safe ? Why is this required ? Azure jars work with 2.7.0 only ?

abhishekdas99: The NativeAzureFileSystem is available in hadoop 2.7. We are not referencing com.microsoft.azure for that.

stagraqubole: This needs to be checked for regression with currently supported versions of hive and presto

abhishekdas99: I have changed the hadoop version back to 2.6 and added only 2.7 version dependency for hadoop-azure module. This is not available in 2.6

rubix-hadoop2/pom.xml
```diff
@@ -27,6 +27,12 @@
             <version>2.6.0</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-azure</artifactId>
+            <version>2.7.0</version>
```
stagraqubole: Is it compatible with hadoop-core 2.6.0? Were you able to access Blobstore with these changes?

abhishekdas99: I had some problem staring a azure cluster. I am planning to test this out and update here

abhishekdas99: I have tested this patch in our hadoop2 cluster with hive query.

vrajat: Is it possible to provide a node bootstrap that Qubole Presto QA can test with ? Presto support is paramount.

abhishekdas99: Yes. We can provide a node bootstrap to test this.

abhishekdas99: I have tested this in a presto azure cluster as well.

rubix-presto/src/main/java/com/qubole/rubix/presto/CachingPrestoNativeAzureFileSystem.java
```diff
@@ -0,0 +1,63 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License. See accompanying LICENSE file.
+ */
+package com.qubole.rubix.presto;
+
+import com.qubole.rubix.core.CachingFileSystem;
+import org.apache.commons.logging.Log;
```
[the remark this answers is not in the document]
abhishekdas99: I agree. I am also not fond of this. We need to decouple FileSystem class from the cluster manager by making cluster manager as configuration. In that case, the filesystem class can be used for different types of cluster manager and we won't have any duplication.

rubix-hadoop2/pom.xml
abhishekdas99: Changed scope to provided . Tested with qds presto.

PR #93, opened 2018-01-18
fix: dev: Assigning a non-zero weight for each of the files in the cache to start eviction

abhishekdas99:
as per guava source code
https://github.com/google/guava/blob/master/guava/src/com/google/common/cache/LocalCache.java#L2623

the weight of individual items needs to be a positive integer to have that item eligible for eviction.


abhishekdas99:
Closed by https://github.com/qubole/rubix/commit/c87a144e10a6e383102ddf479ad8c70217b71d81

PR #95, opened 2018-02-13
Asynchronous cache warm up

abhishekdas99:


vrajat:
My main feedback is on the thread infrastructure. If I understand correctly, requests are put into `processQueue`. A scheduled thread wakes and moves items from `processedQueue` to `rangeMap`. Another scheduled thread will wake up and process `rangeMap`. 

Instead is it possible to executor pool which also gives a queue. Items are put into the queue and threads process these items. The size of the executor pool can be configurable. This will simplify the code and make it more testable. Also a unit test for this part will improve confidence.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -128,6 +133,7 @@ public BookKeeper(Configuration conf)
             throw new TException(e);
         }
         endBlock = setCorrectEndBlock(endBlock, fileLength, remotePath);
+        log.info(" Start Block : " + startBlock + " End Block : " + endBlock);
```
stagraqubole: there will too many of these requests and logging so much data might cause problems. If not perf issues, it might cause undue pressure of logging system that important logs might get delayed from being flushed

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -176,8 +176,11 @@ public void run()
                     throw new Exception("Could not create BookKeeper Client " + Throwables.getStackTraceAsString(e));
                 }
 
-                if (!bookKeeperClient.readData(remotePath, offset, readLength, header.getFileSize(), header.getLastModified(), header.getClusterType())) {
-                    throw new Exception("Could not cache data required by non-local node");
+                if (!CacheConfig.isParallelWarmupEnabled(conf)) {
```
stagraqubole: if isParallelWarmupEnabled is true, then arent we going ahead with reading the local file without ensuring that the data is present or not?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RemoteFetchProcessor.java
```diff
@@ -0,0 +1,201 @@
...
+  protected void runOneIteration() throws Exception
+  {
+    if (!processQueue.isEmpty()) {
+      FetchRequest request = processQueue.remove();
+      log.info("Processing Request : RemotePath - " + request.getRemotePath() + " Offset : " +
+              request.getOffset() + " Length : " + request.getLength());
+
+      if (!rangeMap.containsKey(request.getRemotePath())) {
+        RangeSet<Long> rangeSet = TreeRangeSet.create();
+        rangeMap.putIfAbsent(request.getRemotePath(), rangeSet);
```
stagraqubole: I think there could be a race between runOneIteration and processRemoteFetchRequest:
- runOneIteration creates empty entry in rangeMap
- Meanwhile processRemoteFetchRequest runs and processes that entry and removes it too
- runOneIteration continues and tries to read the non existent entry

All the more reason to get rid of multiple scheduled services competing for same variables

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -384,37 +394,75 @@ else if (isCached.get(idx).getLocation() == Location.CACHED) {
...
+                            NonLocalRequestChain nonLocalRequestChain =
+                                new NonLocalRequestChain(remoteLocation, fileSize, lastModified,
+                                  conf, remoteFileSystem, remotePath, clusterType.ordinal(), strictMode,
+                                  statistics, nextReadBlock, endBlock);
+                            nonLocalAsyncRequests.put(remoteLocation, nonLocalRequestChain);
+                        }
+                        nonLocalAsyncRequests.get(remoteLocation).addReadRequest(readRequest);
+                        if (!nonLocalAsyncRequests.get(remoteLocation).getDirectReadRequestQueue().isEmpty()) {
+                            if (directReadRequestChain == null) {
+                                nonLocalAsyncRequests.get(remoteLocation).getDirectReadRequestQueue().remove();
```
stagraqubole: Do we need a DirectReadRequestQueue in NonLocalRequestChain? From this logic it seems there can never be more than one entry in that Queue

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
stagraqubole: nvm, looks like we we never get here if isParallelWarmupEnabled is true.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RemoteFetchProcessor.java
abhishekdas99: I agree with that. As per @vrajat 's suggestion I am working on fixing this.

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
abhishekdas99: Removed that queue logic. Also found out one more bug here where the cache status was not returned for the desired block in NonLocalRequestChain. Will be relying on a boolean variable here to figure out whether to need a direct read request or not.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
abhishekdas99: Will make it debug log.

abhishekdas99:
Updated the PR based on feedback.

vrajat:
A couple of other changes from an offline conversation:
- Split RemoteFetchProcessor to its own class. 
- Look into other examples of Throttling and Batching queue internal to Qubole.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RemoteFetchProcessor.java
```diff
@@ -0,0 +1,207 @@
...
+        requestChain.addReadRequest(request);
+      }
+      it.remove();
+
+      log.info("Request added for file: " + requestChain.getRemotePath() + " Number of Requests : " +
+              requestChain.getReadRequests().size());
+      readRequestChainList.add(requestChain);
+    }
+
+    processDownloadRequests(readRequestChainList);
```
vrajat: Is the scheduler blocked until all the tasks are done ? Since the code calls `futures.get` ? Just For My Information. Also I am assuming you have a SchedulerThread for Range -> RemoteFetch to improve chances of batching. You could have directly put into FetchRequest but then there will be no batching.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RemoteFetchProcessor.java
```diff
@@ -0,0 +1,207 @@
...
+    // Till the queue is not empty or there are no more requests which came in before the configured delay time
+    // we are going to collect the requests and process them
+    while (!processQueue.isEmpty()) {
+      FetchRequest request = processQueue.peek();
+      if (currentTime - request.getRequestedTime() < this.requestProcessDelay) {
+        break;
+      }
+
+      if (!rangeMap.containsKey(request.getRemotePath())) {
+        RangeSet<Long> rangeSet = TreeRangeSet.create();
```
vrajat: Is this the code snippet that eventually combines ranges/batches reads ? Can you add unit tests for these two ? What will help is a list of requests and tests that show how they look after they are batched. This will also force you to move this code - while loop - to a separate function.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RemoteFetchProcessor.java
```diff
@@ -0,0 +1,207 @@
...
+    final List<FileDownloadRequestChain> readRequestChainList = new ArrayList<FileDownloadRequestChain>();
+
+    for (Iterator<Map.Entry<String, RangeSet<Long>>> it = rangeMap.entrySet().iterator(); it.hasNext(); ) {
+      Map.Entry<String, RangeSet<Long>> entry = it.next();
+      Path path = new Path(entry.getKey());
+      FileSystem fs = path.getFileSystem(conf);
+      fs.initialize(path.toUri(), conf);
+      FSDataInputStream inputStream = fs.open(path);
+      String localPath = CacheConfig.getLocalPath(entry.getKey(), conf);
+      log.info("Processing Request for File : " + path.toString() + " LocalFile : " + localPath);
```
vrajat: Similar to other comments, will this line also create too much log volume ?

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -384,37 +389,74 @@ else if (isCached.get(idx).getLocation() == Location.CACHED) {
             else {
                 if (isCached.get(idx).getLocation() == Location.NON_LOCAL) {
                     String remoteLocation = isCached.get(idx).getRemoteLocation();
-                    log.debug(String.format("Sending block %d to NonLocalReadRequestChain to node : %s", blockNum, remoteLocation));
-                    if (!nonLocalRequests.containsKey(remoteLocation)) {
-                        NonLocalReadRequestChain nonLocalReadRequestChain =
-                            new NonLocalReadRequestChain(remoteLocation, fileSize, lastModified, conf,
-                                remoteFileSystem, remotePath, clusterType.ordinal(), strictMode, statistics);
-                        nonLocalRequests.put(remoteLocation, nonLocalReadRequestChain);
+
```
vrajat: This is a huge function. Refactor is probably out of scope of this PR. Is there a way we can avoid adding to the number of things it does ? I'll try to understand what is going on here to help break it up.

rubix-tests/src/test/java/com/qubole/rubix/tests/TestCachingInputStream.java
```diff
@@ -26,6 +26,7 @@
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FileSystem;
```
vrajat: Are the changes in both the test files mainly to make the current tests to work ? Am I correct to assume that there are no new unit tests yet for this feature ?

abhishekdas99: Yes.. I have not added any new test cases for this feature.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RemoteFetchProcessor.java
```diff
@@ -0,0 +1,211 @@
...
+    fetchRequestMap.clear();
+  }
+
+  @Override
+  protected Scheduler scheduler()
+  {
+    return Scheduler.newFixedDelaySchedule(processThreadInitalDelay, processThreadInterval, TimeUnit.MILLISECONDS);
+  }
+
+  private void processRemoteFetchRequest(ConcurrentMap<String, FileMetadataRequest> fetchRequestMap,
```
stagraqubole: Can we rename fetchRequestMap? Probably to fileMetadataMap, current name threw me off

abhishekdas99:
The PR is ready for another review. Changes done since the last review:
1. Added unit tests
2. Broken RemoteFetchProcessor class into two class and divided the responsibility. RemoteFetchProcessor takes the request out of the queue and merges the requests. FileDownloader then creates the request chain list from the merged requests and processes the download requests.

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -384,37 +385,74 @@ else if (isCached.get(idx).getLocation() == Location.CACHED) {
             else {
                 if (isCached.get(idx).getLocation() == Location.NON_LOCAL) {
```
vrajat: Is it possible to split outall Non-local code to a new class ? It feels like this code needs an interface and derived classes. However, I cannot be sure with the current state. Also as a novice in this code, I am unable to effectively review this part.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RemoteFetchProcessor.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  protected ConcurrentMap<String, DownloadRequestContext> mergeRequests(long currentTime)
+  {
+    // Till the queue is not empty or there are no more requests which came in before the configured delay time
+    // we are going to collect the requests and process them
+    ConcurrentMap<String, DownloadRequestContext> contextMap = new ConcurrentHashMap<String, DownloadRequestContext>();
+    while (!processQueue.isEmpty()) {
+      FetchRequest request = processQueue.peek();
+      if (currentTime - request.getRequestedTime() < this.requestProcessDelay) {
```
stagraqubole: I think this can be further optimized. Do we really want to keep the request on the queue even though it could be merged to the existing DownloadRequestContext? We could rather make an exception in such a case and pick this FetchRequest if it could be merged in one of the DownloadRequess

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RemoteFetchProcessor.java
```diff
@@ -0,0 +1,114 @@
...
+        break;
+      }
+
+      if (!contextMap.containsKey(request.getRemotePath())) {
+        DownloadRequestContext context = new DownloadRequestContext(request.getRemotePath(), request.getFileSize(),
+            request.getLastModified());
+        contextMap.putIfAbsent(request.getRemotePath(), context);
+      }
+      contextMap.get(request.getRemotePath()).addDownloadRange(request.getOffset(),
```
stagraqubole: This could cause corruption of data (probably for a short interval only) in case a file is modified at source and we have some FetchRequests for previous version and a few for new one. We should drop the DownloadRequest if we find a mismatch in modifiedTime across fetchRequests

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloader.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  protected void processDownloadRequests(List<FileDownloadRequestChain> readRequestChainList)
+  {
+    int sizeRead = 0;
+    List<Future<Integer>> futures = new ArrayList<Future<Integer>>();
+
+    for (FileDownloadRequestChain requestChain : readRequestChainList) {
+      requestChain.lock();
+      Future<Integer> result = processService.submit(requestChain);
```
stagraqubole: We need to think about de-duping requests here as well. There is a high chance that this executor is overwhelmed and the queue contains lot of repeated DownloadRequest-ranges. If not doing in this PR, please open a issue to track it

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RemoteFetchProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+          request.getLastModified());
+      if (!contextMap.containsKey(request.getRemotePath())) {
+        contextMap.putIfAbsent(request.getRemotePath(), context);
+      }
+      else {
+        // This takes care of the case where the last modfied time of a file in the request is not matching
+        // with the same of other requests. We will take the latest modified time as a source of truth.
+        // If the last modfied time in context is less than that of current request, we will remove it from the map
+        // Else we will ignode this request.
+        if (contextMap.get(request.getRemotePath()).getLastModifiedTime() < request.getLastModified()) {
```
abhishekdas99: @stagraqubole does this look good in terms of what you were looking for ?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloader.java
abhishekdas99: Yes we will raise another PR as an enhancement for this

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RemoteFetchProcessor.java
abhishekdas99: Can we take this an enhancement and add that logic as a separate PR ?

stagraqubole: we can take it as enhancement but do open a jira to track it

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/RemoteFetchProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+          contextMap.remove(request.getLastModified());
+          contextMap.putIfAbsent(request.getRemotePath(), context);
+        }
+        else if (contextMap.get(request.getRemotePath()).getLastModifiedTime() < request.getLastModified()) {
```
stagraqubole: this is same as the if condition above, you should change to '>'

PR #99, opened 2018-03-13
Added framework to support metrics for bookkeeper

abhishekdas99:


rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -69,6 +69,9 @@
     private static String diskReadBufferSizeConf = "hadoop.cache.data.disk.read.buffer.size";
     public static String socketReadTimeOutConf = "hadoop.cache.network.socket.read.timeout";
     private static String diskMonitorIntervalConf = "hadoop.cache.disk.monitor.interval";
+    private static String enableMeticsConf = "hadoop.cache.data.metrics.enabled";
```
vrajat: Should the namespace be something in rubix ?

pom.xml
```diff
@@ -49,10 +49,12 @@
         <commons-httpclient.version>3.0.1</commons-httpclient.version>
         <dep.hadoop1.version>1.2.0</dep.hadoop1.version>
         <dep.hadoop2.version>2.6.0</dep.hadoop2.version>
+        <dep.dropwizard.version>3.1.0</dep.dropwizard.version>
     </properties>
 
     <modules>
         <module>rubix-spi</module>
+        <module>rubix-common</module>
```
vrajat: Can you point to resources that you used to learn about codahale and come up with this way of adding dropwizard metrics ? Is there a doc in dropwizards-metrics project or did you copy the style from another project ? HS2 probably ?

vrajat: Another generic comment. Can you add a new page in docs to document these metrics ? Similar to http://docs.qubole.com/en/latest/user-guide/presto/presto-system-metrics.html

This can be a basic doc page right now that just list the metrics. We can build it out incrementally. I suggest you start a new section - `Admin Guide`(Right now there are `Rubix` and `Installation Guide` sections). Under that start a sub-section `Metrics`. With this PR, it should contain a list of metrics you've added. No need of descriptions etc right now. Also put a warning that the page is under construction in parentheses.

vrajat: I am used to the pattern suggested http://metrics.dropwizard.io/3.1.0/getting-started/ and in http://metrics.dropwizard.io/3.1.0/getting-started/#reporting-via-jmx

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -68,11 +69,35 @@ public void run()
...
+            public void run()
+            {
+                String shutdownMsg = "Shutting down BookKeeperServer.";
+                log.info(shutdownMsg);
+                try {
+                    MetricsFactory.close();
+                }
+                catch (Exception e) {
+                    log.error("Error in closing MetricsFactory : " + e.getClass().getName() + " "
+                    + e.getMessage(), e);
```
jordanw-bq: Do we have a standard for whether to use `String.format()` over concatenation like this?

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
jordanw-bq: There's also a typo here. `metics` -> `metrics`

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -69,6 +69,9 @@
...
+    private static String meticsClassConf = "hadoop.cache.data.metrics.class";
```
jordanw-bq: Typo here as well.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -69,6 +69,9 @@
...
+    private static String metricsRepotersConf = "hadoop.cache.data.metrics.reporters";
```
jordanw-bq: Should be `metricsReporters`

jordanw-bq:
The new classes being added here only have a 2 space indent. We should match the indentation of the rest of the project (4 spaces)

abhishekdas99:
We are planning to use 2 space indent. We have to go back and change the indent from 4 space to 2 space of the older files.

stagraqubole:
We should also write a script to get the collated view of these metrics across the cluster. Presto provides extracting jvm metrics from all nodes and show them in a single view which makes it easier to track these metrics, something similar for bookkeeper metrics would be really helpful

pom.xml
vrajat: I am confused by the pattern and the need for `Metrics`, `CodahaleMetrics` and `MetricsFactory` class. This is the pattern I've seen and consistently used:

In the main class - for example - `MetricsApplication.java`:

    class Main {
        final MetricRegistry metrics;
        final EventsController eventsController; // This is the object that is instrumented with metrics
    
        Main() {
            metrics = new MetricsRegistry();
            eventsController =
                    new EventsController(metrics);
        }
 
        void run() {
             ....
        }
    }

   
In a class that generates metrics - for example `EventsController.java`
 
    public class EventsController {
        public final Counter networkExceptions;
        public final Counter cloudfrontHits;
        public final Gauge queueSize;
        public final Counter rejectedExecutionExceptions;
        public final Histogram eventSizes;

        public EventsController(MetricRegistry metrics) {
            networkExceptions = metrics.
                counter(name(EventsController.class, "network-exceptions"));
            cloudfrontHits = metrics.
                counter(name(EventsController.class, "cloudfront-hits"));

            queueSize = metrics.
                register(MetricRegistry.name(EventsController.class, "queue-size"),
                (Gauge<Integer>) () -> arrayBlockingQueue.size());

             rejectedExecutionExceptions = metrics.
                counter(name(EventsController.class, "rejected-exceptions"));

            eventSizes = metrics.histogram(name(EventsController.class, "event-sizes"));
        }

This will remove the need for `MetricFactory`, `Metrics` and `Codahale Metrics`. It will also remove the need for registration as that happens right at the beginning as well as locks. AFAIK codahale metrics are thread safe.

The reporter should also be started in the main class. The reporter should filter metrics using a `MetricsFilter`. Example below:

    final SignalFxReporter signalfxReporter = new SignalFxReporter.Builder(
                            metrics,
                            authToken
                    ).setFilter(new MetricFilter())
                     .setMetricMetadata(metricMetadata)
                     .build();

     signalfxReporter.start(60, TimeUnit.SECONDS);

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
vrajat: Let us try to start a trend of having complete PRs. Along with code, a PR should have tests (unit and other types if required) and doc. By this rule, 
- the configuration parameter
- list of metrics
- JMX access

should be documented as part of this feature. Let us use correct names going forward and document them. Then we wont increase the backlog. 
For documentation - configuration and metrics, start a page and create a simple list. We can build it out as we go.

pom.xml
```diff
@@ -49,10 +49,12 @@
         <commons-httpclient.version>3.0.1</commons-httpclient.version>
         <dep.hadoop1.version>1.2.0</dep.hadoop1.version>
         <dep.hadoop2.version>2.6.0</dep.hadoop2.version>
+        <dep.dropwizard.version>3.1.0</dep.dropwizard.version>
```
vrajat: Is it possible to use the latest ? `v4.0.0`

pom.xml
abhishekdas99: 1. The main point of creating Metrics interface and Codahale Metrics class is to have loose coupling. You can have other metrics implementation and that can be set as you metrics by setting this "hadoop.cache.data.metrics.class" to your implementation. 

In the implementation you have suggested have direct reference of CodahaleMetrics in the BookKeeper module which we want to avoid. 

2. The reporter is initialized in the initialization of CodahaleMetrics which gets invoked from the main class. The main reason of doing this is the reporters are specific to metrics implementation. For example, we have in this case  JmxReporter from com.codahale.metrics.JmxReporter.

rubix-common/src/main/java/com/qubole/rubix/common/MetricsConstant.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Created by Abhishek on 2/15/18.
+ */
+public final class MetricsConstant
+{
+  private MetricsConstant()
+  {
+  }
+
+  public static final String CACHED_READ_BYTES = "cached_read_bytes";
```
vrajat: dotted convention is the standard. So this one should be something like `bookkeeper.cachedReadBytes`

Also there is a helper function - `MetricRegistry.name(QueueManager.class, "jobs", "size")` I am not a fan of metrics with fully qualified class name in them. Too verbose. 

We can standardize on `daemon.metricUnit`

abhishekdas99: Will make the change. So we are going to rename cached_read_bytes to cachedReadBytes.

pom.xml
jordanw-bq: Should we document these on the repository's wiki as well?

abhishekdas99: We are going to update that as well

vrajat:
Superseded by a separate design.

Issue #105, opened 2018-03-16
Move from 4-space to 2-space indents


jordanw-bq:
Most open-source project use 2-space indents for their source files. We want to use this for RubiX as well.

jordanw-bq:
This will include updating our Checkstyle checks.xml to enforce indenting.

abhishekdas99:
As part of this we are going to update the checkstyle and fix all the errors coming because of the change

abhishekdas99:
https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml

jordanw-bq:
@abhishekdas99 To confirm, will we be completely switching to Google's checks?

abhishekdas99:
@stagraqubole was referring to airlift (facebook) code style. We can pick either of them. My preference is to have 2-space indents rather than 4 space indents.

PR #106, opened 2018-03-22
 chg: dev: #105: update Checkstyle configuration based on Airlift's checks

jordanw-bq:
PR for #105

The indentation will move to 2 spaces rather than 4. Otherwise, this style will be similar to that used by the Airlift project.
Also, builds will fail when code does not match this style specification.

### DIFF
--- a/src/checkstyle/checks.xml
+++ b/src/checkstyle/checks.xml
@@ -2,107 +2,191 @@
 <!DOCTYPE module PUBLIC
         "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
         "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+<!--
+    Checkstyle configuration that checks the Google coding conventions from Google Java Style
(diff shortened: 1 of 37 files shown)


src/checkstyle/checks.xml
```diff
@@ -3,106 +3,181 @@
...
     </module>
     <module name="RegexpMultiline">
-        <property name="format" value="\n\n\Z"/>
-        <property name="message" value="Blank line before end of file"/>
+        <property name="format" value="->\s*\{\s+\}" />
+        <property name="message" value="Whitespace inside empty lambda body" />
+    </module>
+    <module name="RegexpSingleline">
+        <property name="format" value="(class|interface) ([a-zA-Z0-9_])+(&lt;.*&gt;)? (extends|implements)" />
+        <property name="message" value="No new line before extends/implements" />
```
abhishekdas99: Can we remove the checkstyle for "extends a class or interface " where the extend goes on a separate line

Issue #107, opened 2018-03-22
Unnecessary empty files are getting created in nodes which is not responsible for those files


abhishekdas99:
Let say we have 3 nodes in the cluster and we have 12 files (File1 to File12). Lets assume consistent hashing algorithm puts 
File1, File4.. File10 in Node 1, 
File2, File5.. File11 in Node 2,
File3, File6.. File12 in Node 3

So we should only see File1, File4.. File10 and the corresponding mdfiles in Node1.

But we are seeing File2, File3 etc files (which are not supposed be in Node1) are getting created with 0 bytes.

Same thing is happening for mdfiles as well.

From master log, we know out of 2 nodes, 10.0.0.240 is assigned to have store/000000 file

```
2018-03-22T20:56:26.229Z        INFO    hive-hive-357   com.qubole.rubix.core.CachingFileSystem BlockLocation s3a://bigdata-sets/perf/data/tpcds/parquet/scale_1000/store/000000 0 138942 10.0.0.240 totalHosts: 2
```

The file and corresponding md file is present in 10.0.0.240 node

```
[ec2-user@ip-10-0-0-240 ~]$ ls -lrt /media/ephemeral0/fcache/bigdata-sets/perf/data/tpcds/parquet/scale_1000/store/000000*
-rw-r--r-- 1 yarn yarn      1 Mar 22 20:56 /media/ephemeral0/fcache/bigdata-sets/perf/data/tpcds/parquet/scale_1000/store/000000_mdfile
-rw-rw-rw- 1 yarn yarn 138942 Mar 22 20:56 /media/ephemeral0/fcache/bigdata-sets/perf/data/tpcds/parquet/scale_1000/store/000000
[ec2-user@ip-10-0-0-240 ~]$

```

But if we look at the other node, we also see the same file although the size is zero
```

[ec2-user@ip-10-0-0-11 ~]$ ls -lrt /media/ephemeral0/fcache/bigdata-sets/perf/data/tpcds/parquet/scale_1000/store/000000*
-rw-rw-rw- 1 root root 0 Mar 22 20:56 /media/ephemeral0/fcache/bigdata-sets/perf/data/tpcds/parquet/scale_1000/store/000000
-rw-r--r-- 1 yarn yarn 0 Mar 22 20:56 /media/ephemeral0/fcache/bigdata-sets/perf/data/tpcds/parquet/scale_1000/store/000000_mdfile
[ec2-user@ip-10-0-0-11 ~]$
```

[...]

PR #108, opened 2018-03-23
fix: dev: #107: Fix creation of unnecessary zero bytes file in the cache

abhishekdas99:
After the fix:

```
[ec2-user@ip-10-0-0-251 ~]$ find /media/ephemeral0/fcache/bigdata-sets/perf/data/tpcds/parquet/scale_1000/store_sales/ -type f -size 0b -print
[ec2-user@ip-10-0-0-251 ~]$   

```

```
[ec2-user@ip-10-0-0-251 ~]$ ls -lrt /media/ephemeral0/fcache/bigdata-sets/perf/data/tpcds/parquet/scale_1000/store_sales/00000*
-rw-r--r-- 1 yarn yarn        7 Mar 22 23:56 /media/ephemeral0/fcache/bigdata-sets/perf/data/tpcds/parquet/scale_1000/store_sales/000003_mdfile
-rw-rw-rw- 1 yarn yarn 57186437 Mar 22 23:56 /media/ephemeral0/fcache/bigdata-sets/perf/data/tpcds/parquet/scale_1000/store_sales/000003
```

### DIFF
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
@@ -137,17 +137,17 @@ public class BookKeeper
             for (long blockNum = startBlock; blockNum < endBlock; blockNum++) {
                 totalRequests++;
                 long split = (blockNum * blockSize) / splitSize;
-                if (md.isBlockCached(blockNum)) {
-                    blockLocations.add(new BlockLocation(Location.CACHED, blockSplits.get(split)));
-                    cachedRequests++;
+                if (!blockSplits.get(split).equalsIgnoreCase(nodeName)) {
+                    blockLocations.add(new BlockLocation(Location.NON_LOCAL, blockSplits.get(split)));
                 }
                 else {
-                    if (blockSplits.get(split).equalsIgnoreCase(nodeName)) {
-                        blockLocations.add(new BlockLocation(Location.LOCAL, blockSplits.get(split)));
-                        remoteRequests++;
+                    if (md.isBlockCached(blockNum)) {
+                        blockLocations.add(new BlockLocation(Location.CACHED, blockSplits.get(split)));
+                        cachedRequests++;
                     }
                     else {
-                        blockLocations.add(new BlockLocation(Location.NON_LOCAL, blockSplits.get(split)));
+                        blockLocations.add(new BlockLocation(Location.LOCAL, blockSplits.get(split)));
+                        remoteRequests++;
                     }
                 }
             }
--- a/rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
+++ b/rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
@@ -34,7 +34,6 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.util.DirectBufferPool;
 
-import java.io.File;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.HashMap;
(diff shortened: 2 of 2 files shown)


rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -133,19 +132,6 @@ private void initialize(Configuration conf, BookKeeperFactory bookKeeperFactory)
     this.blockSize = CacheConfig.getBlockSize(conf);
     this.localPath = CacheConfig.getLocalPath(remotePath, conf);
     this.diskReadBufferSize = CacheConfig.getDiskReadBufferSizeDefault(conf);
-    File file = new File(localPath);
-    if (!file.exists()) {
```
stagraqubole: There were multiple commits in git blame in the removed code so there is some history around this piece. Can there be side effects of removing this? e.g some other class might be expecting file to be already there because of this code or expects right read/write perms set by this code

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -136,17 +136,17 @@ public BookKeeper(Configuration conf)
       for (long blockNum = startBlock; blockNum < endBlock; blockNum++) {
         totalRequests++;
         long split = (blockNum * blockSize) / splitSize;
-        if (md.isBlockCached(blockNum)) {
-          blockLocations.add(new BlockLocation(Location.CACHED, blockSplits.get(split)));
-          cachedRequests++;
+        if (!blockSplits.get(split).equalsIgnoreCase(nodeName)) {
```
stagraqubole: Is this just refactor or is there a functional impact of re-arranging this?

stagraqubole:
Questions:

1. What case were these files getting generated unnecessarily?
2. Does this cause functional issues?

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
abhishekdas99: This code is creating a file if the file doesn't exist. This code is running at the client side (not in bookkeeper) and at that time the client has no idea whether this node should have the file or not. So thats why this piece of code was creating the file unnecessarily. If Rubix decides that it needs cache a file from remote file system, it will go through the RemoteReadRequestChain where we are checking whether the corresponding local file exists or not. If not we are creating the then. So this piece of code is redundant.

This is from RemoteReadRequestChain
```
public Integer call()
      throws IOException
  {
    Thread.currentThread().setName(threadName);
    checkState(isLocked, "Trying to execute Chain without locking");

    if (readRequests.size() == 0) {
      return 0;
    }

    // Issue-53 : Open file with the right permissions
    File file = new File(localFile);
    if (!file.exists()) {
      file.createNewFile();
      file.setWritable(true, false);
      file.setReadable(true, false);
    }
```

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
abhishekdas99: Nope this is not just refactor. In the previous case when we were checking for the cached one (now we are checking for non-local one), md. isBlockCached method will go ahead and create the md file if it doesn't exist. So in this case we were also creating md files for files which won't be residing on that node. 

By checking non-local first we are making sure if the file is non going to reside on this node, we are not calling md.isBlockCached hence not creating the file.

abhishekdas99:
@stagraqubole My replies to your comment will give you the scenarios. Just reiterating here as well,
1. The CachingInputStream code was creating the actual file unnecessarily.
2. When we are checking whether the block is cached or not without knowing whether the md file should be in that node, we were also creating the md files for those files as well. By adding the non-local check ahead of cached check is helping not to create the md files unnecessarily.

abhishekdas99:
@vrajat We have some integration tests but that test won't bring any daemon as a separate jvm.

stagraqubole:
Thanks for the clarifications, lgtm

Issue #110, opened 2018-03-28
The number of bytes downloaded should match with the number of requested bytes.


abhishekdas99:
In the bookkeeper readdata method, the data is downloaded in the cache by RemoteReadRequestChain call method. 
We should verify the number of bytes downloaded with the number of requested bytes.
Based on that we should update the cache metadata

PR #111, opened 2018-03-29
fix: dev: #110: Downloaded byte size should match with the requested one

abhishekdas99:


abhishekdas99:
@stagraqubole I think the data read will be less than the expected when there is something wrong during the read. In theory this can happen in normal RemoteRRC execution as well but primarily this is happening because of multiple threads trying to read data at the same time with fs cache being enabled. So the possibility of this happening in case of normal execution is less.
In normal case, the instance of the remotefile system is created by 

```
public CachingFileSystem()
  {
    try {
      this.fs = getTypeParameterClass().newInstance();
    }
    catch (InstantiationException | IllegalAccessException e) {
      log.error("cannot instantiate base filesystem ", e);
      Throwables.propagate(e);
    }
  }
```
So there won't be resuse of the same instance of filesystem

PR #116, opened 2018-04-04
new: dev: add unit tests for cache config utility methods

jordanw-bq:


rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
```diff
@@ -0,0 +1,327 @@
...
+  {
+    final List<String> dirPrefixList = getDirPrefixList(conf);
+    final int maxDisks = CacheConfig.getCacheMaxDisks(conf);
+
+    for (String dirPrefix : dirPrefixList) {
+      for (int i = 0; i < maxDisks; ++i) {
+        final String cacheParentDir = dirPrefix + i;
+        log.debug("Checking for " + cacheParentDir);
+
+        if (exists(cacheParentDir)) {
```
abhishekdas99: Throw exception if the cacheParentDir doesn't exist. There is no point of continuing as the directory where we are going to cache the files doesn't exist.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  /**
+   * Determine whether data should be cached.
+   *
+   * @param path  The data location path.
+   * @param conf  The current Hadoop configuration.
+   * @return True if caching should be skipped, false otherwise.
+   */
+  public static boolean skipCache(Path path, Configuration conf)
```
abhishekdas99: Lets take String as an argument rather than path to avoid confusion.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -348,9 +350,17 @@ private long setCorrectEndBlock(long endBlock, long fileLength, String remotePat
 
   private static synchronized void initializeCache(final Configuration conf)
   {
+    try {
+      CacheUtil.createCacheDirectories(conf);
+    }
+    catch (FileNotFoundException e) {
+      log.error("Cache directories could not be created", e);
+      System.exit(0);
```
abhishekdas99: We should not be doing System.exit here. Let the caller decide what needs to be done. BookKeeperServer initializes BookKeeper class. We should catch the exception there and do the needful.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -70,7 +72,14 @@ public void run()
...
   public static void startServer(Configuration conf)
   {
-    bookKeeper = new BookKeeper(conf);
+    try {
+      bookKeeper = new BookKeeper(conf);
+    }
+    catch (FileNotFoundException e) {
+      log.error("Cache directories could not be created", e);
+      return;
+    }
```
jordanw-bq: @abhishekdas99 The exception thrown when trying to create the cache directories will now be caught when trying to initialize BookKeeper.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
jordanw-bq: Exception is now thrown when `cacheParentDir` does not exist

jordanw-bq:
@abhishekdas99 @vrajat @stagraqubole Is there anything else you'd like me to address on this PR before it gets merged?

rubix-presto/pom.xml
```diff
@@ -32,6 +32,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-hdfs</artifactId>
+            <version>${dep.hadoop2.version}</version>
+            <scope>provided</scope>
```
abhishekdas99: Why this is required ? There is no change in the presto module.

jordanw-bq: My mistake. Thought this was required to resolve a dependency when trying to build after I merged. I'll remove this.

Issue #115, opened 2018-04-03
Write unit tests for rubix-hadoop2 module


abhishekdas99:
This will be similar to TestClusterManager in rubix-presto module

PR #118, opened 2018-04-21
 new: dev: #115: add tests to rubix-hadoop2 module to match rubix-presto

jordanw-bq:


rubix-hadoop2/pom.xml
```diff
@@ -17,10 +17,17 @@
...
             <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-core</artifactId>
+            <artifactId>hadoop-common</artifactId>
+            <version>${dep.hadoop2.version}</version>
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-hdfs</artifactId>
```
abhishekdas99: Where are we using this dependency. Do we need this ?

rubix-hadoop2/pom.xml
```diff
@@ -17,10 +17,17 @@
 
         <dependency>
             <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-core</artifactId>
+            <artifactId>hadoop-common</artifactId>
```
abhishekdas99: Can you provide some comment why we have to change this

jordanw-bq: `hadoop-core` is a hadoop1 dependency, whereas `hadoop-common` is the hadoop2 version of this dependency. Steps such as initializing the Yarn configuration for Hadoop2ClusterManager requires the dependency upgrade.

rubix-hadoop2/pom.xml
jordanw-bq: Due to the `hadoop-core` -> `hadoop-common` dependency update, this is required to provide the `DistributedFileSystem` class for `CachingDistributedFileSystem`.

rubix-hadoop2/pom.xml
abhishekdas99: ok. sounds good

rubix-hadoop2/pom.xml
abhishekdas99: sounds good.

stagraqubole: could the classes from this clash with the installed one? hadoop-common is provided scope while this will be packed in assembly due to compile time dependency.

We need integration tests that can smoke test PRs against presto and hive to answer such questions. Probably something to pick up soon

jordanw-bq: @stagraqubole I agree that integration tests should be an upcoming priority for testing situations like this.

jordanw-bq:
@abhishekdas99 @vrajat @stagraqubole Is there anything else you'd like me to address on this PR before this gets merged?

abhishekdas99:
The code looks good. We are worried about the change in the dependency and the impact it will have during runtime. We have to test this out in a cluster. I will send you a separate email with the instructions.

jordanw-bq:
@abhishekdas99 @vrajat These dependency changes have been successfully tested on a Qubole cluster.

PR #120, opened 2018-04-28
new: dev: add implementation & test to verify providing metrics

jordanw-bq:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -38,6 +39,9 @@
   public static BookKeeper bookKeeper;
   public static BookKeeperService.Processor processor;
 
+  // Registry for gathering & storing necessary metrics
+  private static MetricRegistry metrics = new MetricRegistry();
```
vrajat: Can you make this a member variable that is initialized in the main function or in startServer ?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -62,6 +65,8 @@
  */
 public class BookKeeper implements com.qubole.rubix.spi.BookKeeperService.Iface
 {
+  public static final String METRIC_TOTAL_BLOCK_HITS = name(BookKeeper.class, "total-block-hits");
```
vrajat: There are many suggestions for naming conventions on the internet. This is the best I found: https://docs.signalfx.com/en/latest/best-practices/naming-conventions.html Lets follow that. In this case the metric name will be `rubix.bookkeeper.total_block_hits`. Let's debate this in the design doc.

vrajat: This is better: `rubix.bookkeeper.local_cache.count`

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
abhishekdas99: Should we also have a flag to enable metrics ? Just an implementation detail so that we can turn metrics off if anything goes wrong

vrajat:
A flag to report is definitely required but that is out of scope of this
PR. A flag for instrumentation is a good idea but I am concerned that there
will be too many if conditions all over the code.

vrajat:
@jordanw-bq Can you also list out how you have tested it beyond unit tests ? We have to make sure this does not break anything during run time. Tests are required on both Qubole clusters (P0) and EMR Presto (P1) but we can punt on EMR for now.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
jordanw-bq: Metric name & accompanying variables have been renamed

vrajat:
@jordanw-bq I have approved the changes. Code looks good. Please make sure you test it on a live Presto cluster due to the dependency changes and also get approval from @abhishekdas99 or @stagraqubole .

jordanw-bq:
@vrajat @abhishekdas99 These dependency changes have been successfully tested on a Qubole cluster.

vrajat:
I am going to merge this one. There is a slight drop in coverage and that can be ignored.

Issue #113, opened 2018-03-30
Cache is evicting files even when the maximum size hasn't reached


abhishekdas99:
Rubix uses Guava cache internally to figure out which files to evict. Currently it is evicting files even though the maximum size of the cache hasn't reached. This is very random in nature. This is affecting performance of the jobs using Rubix as its expecting the data to be already cached.

abhishekdas99:
Found this. 
https://stackoverflow.com/questions/10236057/guava-cache-eviction-policy

This is happening because
`Guava caches are segmented into concurrencyLevel different hash tables to allow multiple concurrent reads and writes. The default concurrencyLevel is 4. Basically, if your maximumSize is set to 100, then that actually only results in each of the four segments getting a maximumSize of 25. This is why the maximumSize documentation states:`

abhishekdas99:
The theory is we put the disk space  dummy key in the cache and that item goes in one segment. So if we have 4 segments, the cache should start evicting data when its 1/4th full. We should run experiment and validate this theory. One its validated, we can make changes to put total size/ num segments as weight for the dummy key

abhishekdas99:
Fixed by 61bba2afdf5eb300acce6051bcf3420e2e8472de

PR #125, opened 2018-05-02
fix: dev: #113: Weight for disk space used should consider the concurrency level of guava cache

abhishekdas99:


vrajat:
Can you provide any details on manual testing you have done to ensure that this solves the cache eviction problem ?

abhishekdas99:
@stagraqubole @vrajat  Can anyone take a look at the PR ?

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -105,6 +106,7 @@
   private static final int DEFAULT_SERVER_MAX_THREADS = Integer.MAX_VALUE;
   private static final int DEFAULT_SERVER_PORT = 8899;
   private static final int DEFAULT_SOCKET_READ_TIMEOUT = 30000; // ms
+  private static final int DEFAULT_CACHE_CONCURRENCY_LEVEL = 4;
```
vrajat: A way to turn off the change is required. Can the default value be 1 ? That way the change is not _rolled out_. Based on an account feature we can set it to 4. For any other value, a bootstrap can be used.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -290,6 +292,16 @@ public static boolean isParallelWarmupEnabled(Configuration conf)
     return conf.getBoolean(KEY_PARALLEL_WARMUP, DEFAULT_PARALLEL_WARMUP);
   }
 
+  public static int getCacheConcurrencyLevel(Configuration conf)
```
vrajat: Is a unit test required ? I think @jordanw-bq added unit tests for this file.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
abhishekdas99: The value has been 4. We are not changing the default value. We just changed the weight for the dummy key.
The default value can be found here.
https://github.com/google/guava/blob/master/guava/src/com/google/common/cache/CacheBuilder.java#L159

vrajat: Hmm. I am out of my depth here. I interpreted the change as:
- Before _getWeight_ function returned 100 but now returns 25
- There maybe unintended consequences due to the change in weight for the same _usedMb_. 
- If that is true, then a roll out is requored.

abhishekdas99:
This PR is obsolete because of this commit 61bba2afdf5eb300acce6051bcf3420e2e8472de. We have got rid off the DiskMonitorService

PR #133, opened 2018-05-03
new:usr:Clean up README. Open issues to move most of the contents to docs

vrajat:
Initially README was the only source of documentation. We have recently
created a doc site on readthedocs.org. Therefore we can now organize the docs
in a better way there. As a first step, this commit cleans up README and opens
the following issues:

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,118 +7,28 @@ RubiX is a light-weight data caching framework that can be used by Big-Data engi
 RubiX can be extended to support any engine that accesses data in cloud stores using Hadoop FileSystem interface via plugins. 
 Using the same plugins, RubiX can also be extended to be used with any cloud store
 
-### Usecase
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,118 +7,28 @@ RubiX is a light-weight data caching framework that can be used by Big-Data engi
 RubiX can be extended to support any engine that accesses data in cloud stores using Hadoop FileSystem interface via plugins. 
 Using the same plugins, RubiX can also be extended to be used with any cloud store
 
-### Usecase
+Check the [User and Developer manual]
+## Usecase
 
 RubiX provides disk or in-memory caching of data, which would otherwise be accessed over network when it resides in cloud store,
```
abhishekdas99: should we say in-memory ? I know we use page cache but in terms of "in-memory" people might think something else

PR #138, opened 2018-05-10
new: dev: increase coverage for rubix-presto module

jordanw-bq:
Only test classes have been introduced, so testing on a Qubole cluster should not be required.


jordanw-bq:
Confirmed that the build passes when built using JDK 8

vrajat:
I'll let others comment on the change to Java8

stagraqubole:
can we test hive with java7 and presto with java8?

Issue #124, opened 2018-04-30
Add health metric for RubiX master and send to Datadog


dannykBQ:
Rubix Master requires a health check to allow setting up an alert to determine if the RubiX is live on that node. This will be sent to StatsD in order to be subsequently sent to Datadog for triggering alerts.
https://bitquill.atlassian.net/browse/QUB-58
https://bitquill.atlassian.net/browse/QUB-47

PR #139, opened 2018-05-10
 new: dev: #124: report liveness metric for BookKeeper daemon

jordanw-bq:
- Changes have been tested on a Qubole cluster

### DIFF
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
@@ -29,6 +29,7 @@ import org.apache.thrift.shaded.transport.TServerSocket;
 import org.apache.thrift.shaded.transport.TServerTransport;
 import org.apache.thrift.shaded.transport.TTransportException;
 
+import java.io.FileNotFoundException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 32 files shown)


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -118,4 +127,40 @@ public static boolean isServerUp()
 
     return false;
   }
+
+  /**
+   * Starts a new thread to report the liveness metric at a fixed rate.
+   */
+  private static void scheduleLivenessMetric()
```
vrajat: I dont think you do not need these classes or the scheduled executor. You only need to register the gauge in the constructor. The statsdreporter (or any reporter) will call a get value regularly. You've already implemented the getValue function to return 1. Also the fact that you had to put the null condition in `registerLivenessCheck` is also a clue that you want to do something only once. 

So the change required is to move the gauge registration (Lines 155-162) to `startServer`

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -118,4 +127,40 @@ public static boolean isServerUp()
...
+
+    private void registerLivenessCheck()
+    {
+      if (metrics.getGauges().get(METRIC_BOOKKEEPER_LIVENESS_CHECK) == null) {
+        metrics.register(METRIC_BOOKKEEPER_LIVENESS_CHECK, new Gauge<Integer>()
+        {
+          @Override
+          public Integer getValue()
+          {
+            return 1;
```
vrajat: This is sufficient for now. I want to have a health registry for all components in the bookkeeper like the server and disk monitor. This function should check the health of all the components and then return 1. I'll file an issue for this.

jordanw-bq:
@vrajat This implementation was intended just for registering the metric. We have another task for reporting that to Datadog via the StatsDReporter (QUB-47)

vrajat:
Ack on the StatsdReporter. Can you also comment on the need for scheduler executor service ?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
jordanw-bq: Carry-over from initially implementing it as a counter. I agree with your notes, and will update the gauge resgistration.

jordanw-bq: Gauge registration has been updated.

PR #142, opened 2018-05-15
fix: usr: ISS-141: Throwing exception only when none of the parent directories exists

abhishekdas99:


vrajat:
I am assuming this is a regression from a refactor. @jordanw-bq / @abhishekdas99 Can one of you also add a unit test ? Or do we need product tests to test these. If yes, can you please file a new issue and label it as product-test (create a new one if required).

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
```diff
@@ -48,16 +48,20 @@ public static void createCacheDirectories(Configuration conf) throws FileNotFoun
...
         log.debug("Checking for " + cacheParentDir);
 
         if (exists(cacheParentDir)) {
+          parentDirectoryExists = true;
           final String cacheDirPath = cacheParentDir + "/" + CacheConfig.getCacheDataDirSuffix(conf);
           createCacheDirectory(cacheDirPath);
         }
         else {
-          throw new FileNotFoundException(String.format("Cache parent directory %s does not exist", cacheParentDir));
+          if (!parentDirectoryExists) {
```
stagraqubole: 1. Shouldnt this be outside the inner loop?

2. Better would be move this out of outer loop and throw exception only if none of the dirPrefix+i directories exist.

jordanw-bq:
Since I don't have permissions to push to the branch to update this PR, here is a .patch file containing the changes:
[cacheDirUnitTest.txt](https://github.com/qubole/rubix/files/2015008/cacheDirUnitTest.txt)
(GitHub only accepts certain file formats, so I've changed the extension)

vrajat:
@jordanw-bq Can you open a new pull request ?

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
abhishekdas99: Yes we can move it out. I was confused with the multiple values of dirPrefix. My thought process was we should be able to create atleast one directory with each of the dirPrefix(s).

stagraqubole: Even for that this needs to be outside inner loop.

Anyhow, It would be better to completely fails things only if we cannot find any place to cache so keeping it outside the outerloop.

abhishekdas99:
Added unite tests.

PR #144, opened 2018-05-18
Send registered metrics to StatsD via StatsDReporter

jordanw-bq:
- Passed smoke test on Qubole cluster
- Currently verifying that metrics are sent to DogStatsD on Qubole cluster using default config settings (localhost:8125)

### DIFF
--- a/rubix-bookkeeper/pom.xml
+++ b/rubix-bookkeeper/pom.xml
@@ -112,12 +112,18 @@
                             <artifactSet>
                                 <includes>
                                     <include>com.google.guava:guava</include>
+                                    <include>com.readytalk:metrics-statsd-common</include>
+                                    <include>com.readytalk:metrics3-statsd</include>
                                 </includes>
                             </artifactSet>
                             <relocations>
                                 <relocation>
                                     <pattern>com.google</pattern>
                                     <shadedPattern>com.google.shaded</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>com.readytalk</pattern>
+                                    <shadedPattern>com.readytalk.shaded</shadedPattern>
                                 </relocation>
                             </relocations>
 
--- a/rubix-bookkeeper/pom.xml
+++ b/rubix-bookkeeper/pom.xml
@@ -69,7 +69,6 @@
             <groupId>com.readytalk</groupId>
             <artifactId>metrics3-statsd</artifactId>
             <version>${dep.metrics-statsd.version}</version>
-            <!--<scope>compile</scope>-->
         </dependency>
 
         <dependency>
--- a/rubix-bookkeeper/pom.xml
+++ b/rubix-bookkeeper/pom.xml
@@ -66,6 +66,19 @@
         </dependency>
 
         <dependency>
+            <groupId>com.readytalk</groupId>
+            <artifactId>metrics3-statsd</artifactId>
+            <version>${dep.metrics-statsd.version}</version>
+            <!--<scope>compile</scope>-->
+        </dependency>
+
+        <dependency>
+            <groupId>com.readytalk</groupId>
+            <artifactId>metrics-statsd-common</artifactId>
+            <version>${dep.metrics-statsd.version}</version>
+        </dependency>
+
+        <dependency>
             <groupId>org.testng</groupId>
             <artifactId>testng</artifactId>
             <version>${dep.testng.version}</version>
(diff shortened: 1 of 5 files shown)


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -113,8 +116,14 @@ public static void startServer(Configuration conf, MetricRegistry metricsRegistr
   /**
    * Register desired metrics.
    */
-  private static void registerMetrics()
+  private static void registerMetrics(Configuration conf)
   {
+    if (CacheConfig.isOnMaster(conf)) {
```
jordanw-bq: @vrajat Added config flag to report to StatsD only from master node

vrajat:
Can you confirm that you have tested on a live cluster ? I looked at the dependencies and they seemed benign.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
vrajat: We need one more config to also control whether statsd runs on the master as well. So the if condition needs to be:

    if (CacheConfig.isOnMaster(conf) && CacheConfig.startStatsDReporter(conf)) {

Actually I suggest the following to be future proof:

    if ((CacheConfig.isOnMaster(conf) && CacheConfig.statsdReportOnMaster(conf)) 
       || (!CacheConfig.isOnMaster(conf) && CacheConfig.statsdReportOnWorker(conf))) {

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -113,8 +116,14 @@ public static void startServer(Configuration conf, MetricRegistry metricsRegistr
...
+      StatsDReporter.forRegistry(metrics)
```
vrajat: Can you assert that values are not null ? What will happen if bad values are given ? Does it show up in logs ?

jordanw-bq:
@vrajat These changes have been tested on a live cluster.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
jordanw-bq: What should the default values for these flags be? My thoughts:
```
.isOnMaster -> false (1 master vs. multiple workers)
.reportOnMaster -> true
.reportOnWorker -> false
```

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
jordanw-bq: Which values? For `metrics` and `conf`?

vrajat: if the value of host or port is wrong.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
vrajat: `reportOnMaster` should also be `false`. We will roll it out gradually through QDS cluster configuration.

jordanw-bq: - Flags added to CacheConfig
- Unit tests added for verifying metrics are reported/not reported based on flag combinations
- Flag combinations tested on Qubole cluster

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
jordanw-bq: These values are access through the configuration, so there will always be default values. If the host/port is invalid, StatsDReporter will log an error.

rubix-bookkeeper/pom.xml
```diff
@@ -112,13 +112,19 @@
                             <artifactSet>
                                 <includes>
                                     <include>com.google.guava:guava</include>
+                                    <include>com.readytalk:metrics-statsd-common</include>
+                                    <include>com.readytalk:metrics3-statsd</include>
```
jordanw-bq: @vrajat StatsDReporter dependencies are now being shaded. 
This has also been tested on a Qubole cluster.

Issue #154, opened 2018-06-13
Make ClusterManager pluggable


abhishekdas99:
Currently we have a tight coupling between RubiX related filesystems and cluster managers. Every FileSystems in RubiX hardcodes the cluster manager it wants to use. We want to make it pluggable so that users can opt for plug in their own cluster manager implementation

PR #156, opened 2018-06-14
new: dev: #154: Make cluster manager pluggable

abhishekdas99:


rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -113,6 +115,8 @@
   private static final int DEFAULT_WORKER_LIVENESS_EXPIRY = 60000; // ms
   private static final int DEFAULT_WORKER_LIVENESS_METRIC_INITIAL_DELAY = 30000; // ms
   private static final int DEFAULT_WORKER_LIVENESS_METRIC_INTERVAL = 30000; // ms
+  private static final String DEFAULT_PRESTO_CLUSTER_MANAGER = "com.qubole.rubix.presto.PrestoClusterManager";
```
vrajat: Are you missing some files ? I dont see the implementation of these classes.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -313,6 +317,28 @@ public static boolean isParallelWarmupEnabled(Configuration conf)
...
+  }
+
+  public static String getHadoopClusterManager(Configuration conf)
+  {
+    return conf.get(KEY_HADOOP_CLUSTER_MANAGER, DEFAULT_HADOOP_CLUSTER_MANAGER);
+  }
+
+  public static String getClusterManagerClass(Configuration conf, ClusterType clusterType)
+  {
+    switch (clusterType) {
```
vrajat: Can the callers use the right keys ? The enum seems extraneous to me.

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -115,6 +118,33 @@ public void initialize(URI uri, Configuration conf) throws IOException
...
+    String clusterManagerClassName = CacheConfig.getClusterManagerClass(conf, clusterType);
+    log.info("Initializing cluster manager : " + clusterManagerClassName);
+
+    try {
+      Class clusterManagerClass = conf.getClassByName(clusterManagerClassName);
+      Constructor constructor = clusterManagerClass.getConstructor();
+      ClusterManager clusterManager = (ClusterManager) constructor.newInstance();
+
+      clusterManager.initialize(conf);
+      setClusterManager(clusterManager);
```
vrajat: Why is `clusterManager` static ?

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
abhishekdas99: PrestoClusterManager and Hadoop2ClusterManager already exist. We are using reflection to create instances of these classes.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
abhishekdas99: The enum already exists. The caller will decide whether it needs a presto cluster manager or hadoop2 cluster manager

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
abhishekdas99: The global cluster manager is static because it should be set only once. If erroneously someone is trying to set this, it should be ignored or throw exception

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
vrajat: Hmm - the available ways for specialization are:
1. CachingFileSystem -> [PrestoCachingFileSystem, Hadoop2CachingFileSystem]
2. ClusterType
3. ClusterManager -> [PrestoClusterManager, Hadoop2ClusterManager]

What you need is a mapping between 1 & 3. You are using 2. which seems unnecessary.

I prefer the following which is a direct mapping between 1 & 3:

    class CachingFileSystem {
        abstract String getClusterManagerClass();
     }

     class PrestoCachingFileSystem {
        String getClusterManagerClass() { return CacheConfig.getPrestoClusterManager();}
     }

Actually there doesnt seem to be much use of 2. I checked for usages of `getClusterType` and I didnt find anything interesting.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
vrajat: oops - apologies - I should have read code.

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
vrajat: OK. I checked the code. This seems wrong. Probably `final` is a better option but this is outside the scope of this PR.

rubix-hadoop2/src/main/java/com/qubole/rubix/hadoop2/CachingDistributedFileSystem.java
```diff
@@ -32,25 +34,17 @@
...
   {
     return SCHEME;
   }
-
-  private synchronized void initializeClusterManager(Configuration conf)
-  {
-    if (clusterManager != null) {
-      return;
-    }
-    clusterManager = new Hadoop2ClusterManager();
```
vrajat: All these classes have a local copy of the derived class of cluster manager and you are not setting those anymore. Definitely needs to be fixed in all the classes.

abhishekdas99: this clustermanager is variable is defined in base class(CachingFileSystem). I am calling setClusterManager in the initializeClusterManager method in the base class
https://github.com/qubole/rubix/pull/156/files/b97f3d8fa3dbbba754a9a564b38520142ccb841b#diff-a33cf3db53caa949ee05fad1d36c82f2R121

vrajat: There is yet another in this class and other derived classes of `CachingFileSystem`. Check line number 31(new) or 29(old)

vrajat: Or you can remove the one in all the derived classes. They seem unnecessary.

abhishekdas99: Done

abhishekdas99: Done

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
abhishekdas99: In BookKeeper server, we need to get cluster manager to get the nodes. And the cluster manager object is set only once for the server. BookKeeper doesn't have reference of CachingFileSystem (it should not have it). In that case we create the cluster manager object based on the cluster type enum. Thats why I took approach 2. We cannot have [getClusterManagerClass](url) as abstract method then in `CachingFileSystem`.

vrajat: OK. I get it. This goes back to `Bookkeeper.readData`. The clients tell who they are. I prefer starting the rubix server with a particular type but thats a separate debate.

abhishekdas99:
Added unit tests for existing and non-existing ClusterManagers

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeper.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+import static org.testng.Assert.assertTrue;
+
+/**
+ * Created by Abhishek on 6/15/18.
+ */
+public class TestBookKeeper
```
vrajat: This is a good start. 

Can you test Presto & H2 cluster types as well  ? I understand that you cant use the default classes for these types.
One way to get around that is to:

    public void testGetClusterManagerInValidInstance() throws Exception
    {
        ClusterType type = ClusterType.PRESTO_CLUSTER_MANAGER; //or H2
        CacheConfig.setDummyClusterManager(conf, "com.qubole.rubix.core.test.PrestoDummyClusterManager");
        BookKeeper bookKeeper = new CoordinatorBookKeeper(conf, metrics);

        ClusterManager manager = bookKeeper.getClusterManagerInstance(type, conf);
        //assert that the class is of the right type. 
    }

abhishekdas99: Done

PR #159, opened 2018-06-20
Fix #158: Add retry logic to HeartbeatService

### LINKED ISSUE #158: HeartbeatService not starting if master BookKeeper not up
If the BookKeeperServer on the master node in a cluster is not running, the HeartbeatService on a worker node will not be able to connect to it and fail to start.

jordanw-bq:
- Successfully tested on a Qubole cluster

### DIFF
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
@@ -111,9 +111,7 @@ public class WorkerBookKeeper extends BookKeeper
           return bookKeeperFactory.createBookKeeperClient(masterHostname, conf);
         }
         catch (TTransportException e) {
-          System.out.println("Could not start client for heartbeat service: ");
-          e.printStackTrace();
-          log.fatal("Could not start client for heartbeat service", e);
+          log.warn("Could not start client for heartbeat service", e);
         }
 
         failedStarts++;
@@ -129,6 +127,7 @@ public class WorkerBookKeeper extends BookKeeper
         }
       }
 
+      log.fatal("Heartbeat service ran out of retries to connect to the master BookKeeper");
       throw new RuntimeException("Could not start heartbeat service");
     }
 
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
@@ -57,7 +57,7 @@ public class WorkerBookKeeper extends BookKeeper
     final int retryInterval = CacheConfig.getServiceRetryInterval(conf);
     final int maxRetries = CacheConfig.getServiceMaxRetries(conf);
 
-    for (int failedStarts = 0; failedStarts < maxRetries;) {
+    for (int failedStarts = 0; failedStarts < maxRetries; ) {
       try {
         this.heartbeatService = new HeartbeatService(conf);
         heartbeatService.startAsync();
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
@@ -55,10 +55,9 @@ public class WorkerBookKeeper extends BookKeeper
   private void startHeartbeatService(Configuration conf)
   {
     final int retryInterval = CacheConfig.getServiceRetryInterval(conf);
-    final int retries = CacheConfig.getServiceMaxRetries(conf);
+    final int maxRetries = CacheConfig.getServiceMaxRetries(conf);
 
-    int failedStarts = 0;
-    while (failedStarts < retries) {
+    for (int failedStarts = 0; failedStarts < maxRetries;) {
       try {
         this.heartbeatService = new HeartbeatService(conf);
         heartbeatService.startAsync();
@@ -68,14 +67,17 @@ public class WorkerBookKeeper extends BookKeeper
         log.fatal("Could not start client for heartbeat service", e);
       }
 
+      failedStarts++;
+      if (failedStarts == maxRetries) {
+        break;
+      }
+
       try {
         Thread.sleep(retryInterval);
       }
       catch (InterruptedException e) {
-        log.error("Thread interrupted while waiting for HeartbeatService to start", e);
+        Thread.currentThread().interrupt();
       }
-
-      failedStarts++;
     }
 
     throw new RuntimeException("Could not start heartbeat service");
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
@@ -23,6 +23,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.thrift.shaded.TException;
+import org.apache.thrift.shaded.transport.TTransportException;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -53,14 +54,14 @@ public class WorkerBookKeeper extends BookKeeper
    */
   private void startHeartbeatService(Configuration conf)
   {
-    this.heartbeatService = new HeartbeatService(conf);
+    this.heartbeatService = new HeartbeatService(conf, new BookKeeperFactory());
     heartbeatService.startAsync();
   }
 
   /**
    * Class to send a heartbeat to the master node in the cluster.
    */
-  private static class HeartbeatService extends AbstractScheduledService
+  protected static class HeartbeatService extends AbstractScheduledService
   {
     private static Log log = LogFactory.getLog(HeartbeatService.class.getName());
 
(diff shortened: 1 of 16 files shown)


vrajat:
The approach seems reasonable. @abhishekdas99 how does Hadoop solve this ? Similar approach ?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -54,13 +54,33 @@ public void handleHeartbeat(String workerHostname)
    */
   private void startHeartbeatService(Configuration conf)
   {
-    try {
-      this.heartbeatService = new HeartbeatService(conf);
-      heartbeatService.startAsync();
-    }
-    catch (TTransportException e) {
-      log.fatal("Could not start heartbeat service", e);
+    final int retryInterval = CacheConfig.getServiceRetryInterval(conf);
```
abhishekdas99: The main problem here is we are creating the bookkeeper client in the constructor of HeartbeatService. So in case when there is an exception while creating the bookkeeper client, we are catching the exception in `startHeartbeatService`

What we should do is to move the creation of bookkeeper client in the `runOneIteration` method and create the client only once. In that case `runOneIteration` takes care of the retry part. So even if the client doesn't get created at first attempt, it will retry at regular interval defined by `heartbeatInterval`

jordanw-bq: Makes sense. I've moved the creation of the client to `runOneIteration`, and the change is working on QiB.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -108,15 +101,22 @@ protected void startUp()
...
-    protected void runOneIteration() throws TException
+    protected void runOneIteration()
     {
       try {
+        if (bookkeeperClient == null) {
+          this.bookkeeperClient = new BookKeeperFactory().createBookKeeperClient(masterHostname, conf);
+          log.debug(String.format("Connected to master node [%s]", masterHostname));
+        }
         log.debug(String.format("Sending heartbeat to %s", masterHostname));
         bookkeeperClient.handleHeartbeat(InetAddress.getLocalHost().getCanonicalHostName());
```
abhishekdas99: There will be a NPE here if `bookkeeperClient` is null

abhishekdas99: My bad.. You are right. `bookkeeperClient` won't be null here

vrajat:
I understand why you moved the client setup to `runOneIteration`. You are using `runOneIteration` as a defacto retry mechanism. One problem with this approach is that the rubix worker is active even though heartBeatService is not yet up. We should try to avoid this situation. Ideally (some of this specific to QDS), Rubix and the machine should be marked as up if all services are up. Rubix and machine should be marked unhealthy if all services are not up. This includes the heartBeatService. Right now, it again goes back to best effort. 

So I suggest the following change: Move initialization of the client back to constructor using a function.
    
    this.bookKeeperClient = initializeClientWithRetry(...)

Implement the retry login in `initializeClientWithRetry`. This will also help to unit test it. You can mock `new BookKeeperFactory().createBookKeeperClient(masterHostname, conf);` to return errors to unit test the retry logic as well. The fact that it is easier to unit test is a good sign that this is a better design.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -91,13 +86,50 @@ private void startHeartbeatService(Configuration conf)
...
+        }
+
+        try {
+          Thread.sleep(retryInterval);
+        }
+        catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
+        }
+      }
+
```
vrajat: Log a fatal message here.

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestWorkerBookKeeper.java
```diff
@@ -0,0 +1,134 @@
...
+        new RetryingBookkeeperClient(
+            new TSocket("localhost", CacheConfig.getServerPort(conf), CacheConfig.getClientTimeout(conf)),
+            CacheConfig.getMaxRetries(conf)));
+
+    startBookKeeperServer();
+    final WorkerBookKeeper.HeartbeatService heartbeatService = new WorkerBookKeeper.HeartbeatService(conf, bookKeeperFactory);
+  }
+
+  /**
+   * Verify that the heartbeat service no longer attempts to connect once it runs out of retry attempts.
```
vrajat: I am not sure if its possible with mock in Java. Can you test if connection is established after a few failures ? Skip if its too hard.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -91,13 +86,50 @@ private void startHeartbeatService(Configuration conf)
...
+    {
+      final int retryInterval = CacheConfig.getServiceRetryInterval(conf);
+      final int maxRetries = CacheConfig.getServiceMaxRetries(conf);
+
+      for (int failedStarts = 0; failedStarts < maxRetries; ) {
+        try {
+          return bookKeeperFactory.createBookKeeperClient(masterHostname, conf);
+        }
+        catch (TTransportException e) {
+          System.out.println("Could not start client for heartbeat service: ");
```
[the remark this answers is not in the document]
jordanw-bq: Stdout message was leftover from testing. I've removed this & changed the log level for the log statement.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
jordanw-bq: Added a fatal log message.

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestWorkerBookKeeper.java
jordanw-bq: Added unit test for connecting after a few retries (see line 99 `testHeartbeatRetryLogic_connectAfterRetries()`)

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestWorkerBookKeeper.java
```diff
@@ -0,0 +1,189 @@
...
+    startBookKeeperServer();
+    final WorkerBookKeeper.HeartbeatService heartbeatService = new WorkerBookKeeper.HeartbeatService(conf, bookKeeperFactory);
+  }
+
+  /**
+   * Start an instance of the BookKeeper server.
+   *
+   * @throws InterruptedException if the current thread is interrupted while sleeping.
+   */
+  private void startBookKeeperServer() throws InterruptedException
```
abhishekdas99: Lets invoke this method in BeforeMethod

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestWorkerBookKeeper.java
```diff
@@ -0,0 +1,189 @@
...
+    };
+    thread.start();
+  }
+
+  /**
+   * Stop the currently running BookKeeper server instance.
+   *
+   * @throws InterruptedException if the current thread is interrupted while sleeping.
+   */
+  private void stopBookKeeperServer() throws InterruptedException
```
abhishekdas99: Invoke this in AfterMethod

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestWorkerBookKeeper.java
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * Verify that the heartbeat service correctly makes a connection using a BookKeeper client after a number of retries.
+   */
+  @Test
+  public void testHeartbeatRetryLogic_connectAfterRetries()
+  {
+    final int retryInterval = 500;
+    CacheConfig.setServiceRetryInterval(conf, retryInterval);
```
abhishekdas99: as BeforeMethod has already started the server, lets stop it first and then invoke startBookKeeperServerWithDelay

abhishekdas99:
The CI build is failing. Other than that it looks fine. @jordanw-bq can you take care of the CI build issue

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeper.java
```diff
@@ -36,12 +39,16 @@
...
+  private static final Log log = LogFactory.getLog(TestBookKeeper.class);
+
   private MetricRegistry metrics;
   private Configuration conf;
 
   @BeforeMethod
-  public void setUp() throws Exception
+  public void setUp(Method method) throws Exception
   {
+    log.info("Starting test " + method.getName());
```
abhishekdas99: Remove this

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeperMetrics.java
```diff
@@ -52,6 +58,12 @@ public void setUp() throws IOException
     bookKeeper = new CoordinatorBookKeeper(conf, metrics);
   }
 
+  @BeforeMethod
+  public void setUpForTest(Method method)
+  {
+    log.info("Starting test " + method.getName());
```
abhishekdas99: Same here

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeperServer.java
```diff
@@ -62,8 +63,10 @@ public void initializeCacheDirectories() throws IOException
   }
 
   @BeforeMethod
-  public void setUp()
+  public void setUp(Method method)
   {
+    log.info("Starting test " + method.getName());
```
abhishekdas99: Same here

abhishekdas99:
@jordanw-bq Do we need `log.info("Starting test " + method.getName());` . If not can we remove this log

jordanw-bq:
@abhishekdas99 I added those lines to help determine the cause of the Travis build timeout, since I didn't experience it when building locally. Now that the build is passing, they can be (and have been) removed.

abhishekdas99:
LGTM

Issue #157, opened 2018-06-15
Push the consistent hashing index calculation logic in ClusterManager


abhishekdas99:
Currently we use consistent hashing to assign a node for a particular file. This is done when 1. when we need to provide a hint to the scheduler for data locality and 2. to figure out whether a file needs to cached in local node or any remote node.

The logic of calculating the hashing index needs to be centralized and ClusterManager should be responsible for this. We have already made ClusterManager pluggable so that different calculation logic can be implemented.

PR #160, opened 2018-06-26
new: dev: #157: Calculate consistently hashed nodeIndex in ClusterManager

abhishekdas99:


abhishekdas99:
@vrajat Implemented the change. The unit test is very basic though

Issue #162, opened 2018-07-02
Implement consistent hashing logic that takes care of the lost nodes or downscaling of a cluster


abhishekdas99:
```
Node Index table		File Index Table 		File-Node Membership
				(Returned by hashing logic)		
A	0			F1	0			F1	A
B	1			F2	1			F2	B
C	2			F3	2			F3	C
D	3			F4	3			F4	D
```

If we lose Node A and  cluster adds a new node E, there will be 4 valid nodes in the cluster. In the current logic in clusterManager, the getNodes call will return a list containing {B, C, D, E} as A is lost.

So the tables look like

```
Node Index table		File Index Table 		File-Node Membership
				(Returned by hashing logic)		
B	0			F1	0			F1	B
C	1			F2	1			F2	C
D	2			F3	2			F3	D
E	3			F4	3			F4	E

```
In this case for all the files, file-node memberships are going to change which defeats the purpose of consistent hashing

PR #163, opened 2018-07-04
new: usr: #162: Added Hadoop2ClusterManagerV2 implememtation of hadoop cluster manager for correct consistent hashing logic

abhishekdas99:


vrajat:
One big question is composition vs inheritance between the two implementations of Hadoop Cluster Managers. I probably prefer inheritance in this specific case because tests that are valid for V1 is also valid for V2 except for a few scenarios where you want different scenarios. If you choose composition, then you have to repeat all the tests.

abhishekdas99:
@vrajat Yes I thought about it. First the tests are same in terms of checking the condition. The output of the tests are going to be different. For example, for `testUnhealthyNodeCluster_decommissioned`, the old HadoopClusterManager will return 1 node whereas the new implementation is going to return 2 nodes. 
Both the clustermanagers are extending the abstract class and quite a few implementations are different for these two cases such as `initialize`, `getNextRunningNodeIndex` and `getPreviousRunningNodeIndex`. The common methods which can be used as utility methods have been extracted out to a different Utility class. Same logic has been applied to TestClasses as well.

PR #166, opened 2018-07-06
Bookkeeperjvm

kvankayala:
This PR includes test case with bookkeeper and LDS as separate JVMs.


rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,176 @@
...
+import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
+
+import static org.testng.Assert.assertTrue;
+
+public class TestThriftServerJVM extends Configured
+{
+  private static final String testDirectoryPrefix = "/Users/kvankayala/Downloads/TestThriftServerJVM/"; //System.getProperty("java.io.tmpdir") + "/TestThriftServerJVM/";
```
vrajat: Hard Coded value

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,176 @@
...
+  public static void cleanup()
+  {
+    File file = new File(backendFileName);
+    file.delete();
+  }
+
+  @Test
+  public void testJVMCommunication() throws IOException, InterruptedException
+  {
+    log.info("Value of Path " + this.backendFileName);
```
vrajat: Seems like there is too much setup here. Does some of this have to move to a setup method ?

abhishekdas99: I agree we should move make conf as static variable and set some of these configs in beforeclass or beforemethod

abhishekdas99:
The tests are going to fail because there is no rubix jars or the hadoop jars in `/usr/lib/`

abhishekdas99:
For now if the tests are running in you local, can you remove the test , beforeclass, beforemethod etc all these attributes so that the tests don't run.

We need to come back to this once the docker container with hadoop is setup

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
kvankayala: I will move some of the initialization code to setup method

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,176 @@
...
+    int readSize = 1000;
+
+    try {
+      client = bookKeeperFactory.createBookKeeperClient(host, conf);
+
+      result = client.getCacheStatus("file:///" + file.toString(), file.length(), file.lastModified(), 0, lastBlock, 3);
+      assertTrue(result.get(0).getLocation() == Location.LOCAL, "File already cached, before readData call");
+      log.info(" Value of Result : " + result);
+      log.info("Downloading file from path : " + file.toString());
+      dataDownloaded = client.readData("file:///" + file.toString(), 0, readSize, file.length(), file.lastModified(), 3);
```
abhishekdas99: Use the `backendFileName` instead of `"file:///" + file.toString()`

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,176 @@
...
+
+    try {
+      client = bookKeeperFactory.createBookKeeperClient(host, conf);
+
+      result = client.getCacheStatus("file:///" + backendFileName, file.length(), file.lastModified(), 0, lastBlock, 3);
+      assertTrue(result.get(0).getLocation() == Location.LOCAL, "File already cached, before readData call");
+      log.info(" Value of Result : " + result);
+      log.info("Downloading file from path : " + file.toString());
+      dataDownloaded = client.readData("file:///" + backendFileName, 0, readSize, file.length(), file.lastModified(), 3);
+      if (!dataDownloaded) {
```
vrajat: The test should not swallow this. This seems like a condition on which the test should fail.

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,176 @@
...
+        log.info("Failed to read Data from the location");
+      }
+      result = client.getCacheStatus("file:///" + backendFileName, file.length(), file.lastModified(), 0, lastBlock, 3);
+      assertTrue(result.get(0).getLocation() == Location.CACHED, "File not cached properly");
+      log.info(" Value of Result : " + result);
+    }
+    catch (TTransportException ex) {
```
vrajat: These exceptions should not be swallowed ?

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
abhishekdas99: We should fail the test if dataDownloaded is false. It means we have not been able to download the file properly.

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,176 @@
...
+    File file = new File(backendFileName);
+    List<BlockLocation> result;
+    int lastBlock = 4;
+    int readSize = 1000;
+
+    try {
+      client = bookKeeperFactory.createBookKeeperClient(host, conf);
+
+      result = client.getCacheStatus("file:///" + backendFileName, file.length(), file.lastModified(), 0, lastBlock, 3);
+      assertTrue(result.get(0).getLocation() == Location.LOCAL, "File already cached, before readData call");
```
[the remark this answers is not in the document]
abhishekdas99: @vrajat We are checking here the initial condition where the file should be in this node (LOCAL for that reason) but have not been downloaded yet. Yes you are right that the file has not been cached yet.

The whole flow is first check whether the file is in the cache (which it should not be) then download the file and then check the cache status.

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
abhishekdas99: We should fail the test if we get any TTransportException as the server is up and running and there should not be any error during communicating the server

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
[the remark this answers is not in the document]
vrajat: Got it.

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
kvankayala: I have included assert statements in exceptions

vrajat: I suggest you remove it. If there is an exception, then the test will fail and all the information will be in the test o/p. Also I suggest you add another test that starts a bookkeeper client to some port where bookkeeper is not running and expect an exception. 

Rest of the changes lgtm.

kvankayala: @vrajat I understood your use case now. I will add another test case and will make sure all the comments are addressed properly in the next commit.

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,202 @@
...
+    log.info("Downloading file from path : " + file.toString());
+    dataDownloaded = client.readData("file:///" + backendFileName, 0, readSize, file.length(), file.lastModified(), 3);
+    assertTrue(dataDownloaded == true, "readData() function call failed. File not downloaded properly");
+
+    result = client.getCacheStatus("file:///" + backendFileName, file.length(), file.lastModified(), 0, lastBlock, 3);
+    assertTrue(result.get(0).getLocation() == Location.CACHED, "File not cached properly");
+    log.info(" Value of Result : " + result);
+  }
+
+  @Test(enabled = false)
```
vrajat: Almost there. Check https://github.com/junit-team/junit4/wiki/exception-testing to learn how to test that exception is fired.

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,178 @@
...
+    String host = "localhost";
+    File file = new File(backendFileName);
+    List<BlockLocation> result;
+    int lastBlock = 4;
+    int readSize = 1000;
+
+    RetryingBookkeeperClient client;
+    client = bookKeeperFactory.createBookKeeperClient(host, conf);
+
+    result = client.getCacheStatus("file:///" + backendFileName, file.length(), file.lastModified(), 0, lastBlock, 3);
```
abhishekdas99: Use `backendPath` here instead of `"file:///" + backendFileName`

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,178 @@
...
+    assertTrue(result.get(0).getLocation() == Location.LOCAL, "File already cached, before readData call");
+    log.info(" Value of Result : " + result);
+    log.info("Downloading file from path : " + file.toString());
+    boolean dataDownloaded = client.readData("file:///" + backendFileName, 0, readSize, file.length(), file.lastModified(), 3);
```
abhishekdas99: Use `backendPath` here instead of `"file:///" + backendFileName`

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,178 @@
...
+    assertTrue(dataDownloaded == true, "readData() function call failed. File not downloaded properly");
+
+    result = client.getCacheStatus("file:///" + backendFileName, file.length(), file.lastModified(), 0, lastBlock, 3);
+    assertTrue(result.get(0).getLocation() == Location.CACHED, "File not cached properly");
```
abhishekdas99: Check for all the blocks

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,178 @@
...
+    File file = new File(backendFileName);
+    List<BlockLocation> result;
+    int lastBlock = 4;
+    int readSize = 1000;
+
+    RetryingBookkeeperClient client;
+    client = bookKeeperFactory.createBookKeeperClient(host, conf);
+
+    result = client.getCacheStatus("file:///" + backendFileName, file.length(), file.lastModified(), 0, lastBlock, 3);
+    assertTrue(result.get(0).getLocation() == Location.LOCAL, "File already cached, before readData call");
```
abhishekdas99: Check for all the blocks

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
kvankayala: Addressed

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
kvankayala: Addressed

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
kvankayala: Done

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
kvankayala: Done

rubix-tests/src/test/java/com/qubole/rubix/tests/TestThriftServerJVM.java
```diff
@@ -0,0 +1,178 @@
...
+
+import static org.testng.Assert.assertTrue;
+
+/**
+ * Created by kvankayala on 10 Jul 2018.
+ * TestThriftServerJVM Class created to test separate JVMs for bookKeeper and LDS servers
+ */
+public class TestThriftServerJVM extends Configured
+{
+  private static final String testDirectoryPrefix = System.getProperty("java.io.tmpdir") + "/TestThriftServerJVM/";
```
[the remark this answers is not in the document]
kvankayala: As we discussed, this issue will not be handled in this PR because the changes that jordan had made were in rubix-bookkeeper module which are currently not available in rubix-tests module

Issue #164, opened 2018-07-04
Register JVM Metrics


jordanw-bq:
As requested by Das, register metrics to report status & performance details for the currently running JVM

PR #171, opened 2018-07-10
new: dev: #164: register JVM metrics and report metrics to JMX

jordanw-bq:
* Changes have been tested on a Qubole cluster


vrajat:
This is good stuff. One doubt I have is that too many metrics will be reported suddenly to statsd and datadog. Does it make sense to have a flag for a category of metrics ? JVM metrics seems like a good category. I am not sure what is the best practice to have flags to turn on certain category of metrics. I am assuming implementation is straightforward using metricsfilter.

jordanw-bq:
@vrajat I agree regarding the number of metrics being reported. I think it makes sense to add flags to  enable/disable different metrics categories. Maybe something like `rubix.metrics.jvm.enabled`?
And yes, the implementation should be straightforward.

jordanw-bq:
@vrajat I've added metrics categories and flags for enabling/disabling them.

jordanw-bq:
@vrajat I've updated the implementation to filter at the reporter level using a custom MetricFilter.

jordanw-bq:
@abhishekdas99 @vrajat Metrics category changes are confirmed to be working in a cluster.

rubix-bookkeeper/pom.xml
```diff
@@ -128,6 +134,7 @@
                                     <include>com.google.guava:guava</include>
                                     <include>com.readytalk:metrics-statsd-common</include>
                                     <include>com.readytalk:metrics3-statsd</include>
+                                    <include>io.dropwizard.metrics:metrics-jvm</include>
```
abhishekdas99: Are we shading the jar here ? If yes, can we put a relocation for this so that we can differentiate this dependency from other dropwizard jar

rubix-common/src/main/java/com/qubole/rubix/common/metrics/BookKeeperMetricsFilter.java
```diff
@@ -0,0 +1,54 @@
...
+{
+  private final Configuration conf;
+
+  public BookKeeperMetricsFilter(Configuration conf)
+  {
+    this.conf = conf;
+  }
+
+  @Override
+  public boolean matches(String name, Metric metric)
```
abhishekdas99: We should move the creation of `whitelist` in the constructor so that we don;t have to do the same thing everytime `matches` is invoked

rubix-common/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                        <phase>test-compile</phase>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
abhishekdas99: Are we shading the `io.dropwizard.metrics` in bookkeeper module. If yes, why aren't we shading in rubix-common ?

rubix-common/src/main/java/com/qubole/rubix/common/metrics/BookKeeperMetricsFilter.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    for (String whitelistedMetric : whitelist) {
+      if (name.startsWith(whitelistedMetric)) {
+        return true;
+      }
+    }
+    return false;
+  }
+}
```
abhishekdas99: We are getting this exception in the test
```
2018-08-02 00:16:13 [DEBUG] metrics.JmxReporter: Unable to unregister gauge
javax.management.InstanceNotFoundException: metrics:name=rubix.bookkeeper.jvm.threads.new.count
	at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.getMBean(DefaultMBeanServerInterceptor.java:1095)
	at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.exclusiveUnregisterMBean(DefaultMBeanServerInterceptor.java:427)
	at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.unregisterMBean(DefaultMBeanServerInterceptor.java:415)
	at com.sun.jmx.mbeanserver.JmxMBeanServer.unregisterMBean(JmxMBeanServer.java:546)
	at com.codahale.metrics.JmxReporter$JmxListener.unregisterMBean(JmxReporter.java:521)
	at com.codahale.metrics.JmxReporter$JmxListener.onGaugeRemoved(JmxReporter.java:543)
	at com.codahale.metrics.MetricRegistry.notifyListenerOfRemovedMetric(MetricRegistry.java:366)
	at com.codahale.metrics.MetricRegistry.onMetricRemoved(MetricRegistry.java:360)
	at com.codahale.metrics.MetricRegistry.remove(MetricRegistry.java:156)
	at com.codahale.metrics.MetricRegistry.removeMatching(MetricRegistry.java:170)
	at com.qubole.rubix.bookkeeper.BookKeeperServer.removeMetrics(BookKeeperServer.java:149)

```

The main reason is we are registering all the metrics (JVM, LIVELINESS) in the registerMetrics while starting the server.
But based on the flag, we are only pushing some metrics in JMX through BookKeeperMetricsFilter. So during unregister, the system is trying to remove some metrics (already registered) and failing because those are not available in JMX. (By default the JMX metrics are not pushed).
We need to handle this scenario.

rubix-common/pom.xml
jordanw-bq: We are shading the `io.dropwizard.metrics` for the `:metrics-jvm` artifact in `rubix-bookkeeper`. Since we're only using `:metrics-core` in `rubix-common` (which is a provided dependency), we won't be shading there.

rubix-bookkeeper/pom.xml
jordanw-bq: Added relocation

jordanw-bq:
New changes are confirmed to be working on a Qubole cluster

jordanw-bq:
@abhishekdas99 I agree. It would be better to include everything we need, rather than depend on an environment, especially one out of our control, to have what we need.
I'll go ahead and make this change, then test with an EMR cluster.

abhishekdas99:
@jordanw-bq yes. Before making the change and testing it, can you please check whether with your current change, is it running fine or not

jordanw-bq:
Changes (including the shading changes) have been partially tested on an EMR cluster to ensure the Rubix daemons start correctly. I'll be smoke testing queries on Monday.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -58,7 +58,7 @@
   private static Log log = LogFactory.getLog(BookKeeperServer.class.getName());
   private static BookKeeperMetrics bookKeeperMetrics;
 
-  protected BookKeeperServer()
+  public BookKeeperServer()
```
abhishekdas99: are we using this constructor outside this bookkeeper package ? If not we should not be making this public. Method without any modifier is accessible within the same package. So we should just use no modifier if its only used in the same package 

This applies to all other places where we are using public modifier

jordanw-bq: It's used in tests like `TestCachingInputStream` in `rubix-tests` for starting the BookKeeperServer, now that I've removed the static modifiers.

https://github.com/qubole/rubix/blob/af9089df037e572d3265366e266589756abc8859/rubix-tests/src/test/java/com/qubole/rubix/tests/TestCachingInputStream.java#L106-L114

abhishekdas99: ok .. makes sense

PR #181, opened 2018-07-19
 Changes made for Continuous Integration process with docker in Travis

kvankayala:
Changed .travis.yml file
Added script.sh
Uses docker image "kvankayala/hadoop_mvn_thrift" for Continuous Integration.


vrajat:
Nice! I could see the testng o/p in travis. I have a few high level questions:

- script.sh is missing cobertura run.
- This is running unit tests. Docker is required for integration tests. Should #166 be merged first ? That is the real test for docker.
- Is it possible to run unit tests and cobertura outside docker and then use docker for integration tests ?
- If you decide to go with dockers for all tests, can you avoid maven download of dependencies ?
- Can you move the docker to qubole account from your personal account ?

kvankayala:
@vrajat 
1) I will add the cobertura to travis.yml
2)   a. Both #166 and #181 are interdependent on each other i.e., #166 validates changes in #181  and vice-versa.
     b. Plan is first merge #181 which provides infrastructure (docker) to validate tests in #166.
     c. Otherwise, I can push #166 changes with #181 in a temporary commit (will remove  #166 changes after validation) and validate both together and later approves them separately.
3) Yes, it could be done. But, this docker container has all the necessary dependencies for rubix testing. So, it is better to execute all the tests together in Docker.
4) I could explore about this. Currently, I feel that maven dependencies download could be avoided in docker. But, We would like to leave it for improvements once this initial jvm and docker stuff works good.
5) I will move the container to quboleinc

abhishekdas99:
Lets add the DOCKERFILE in the repo for the sake of completeness.

Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+    4A6AC5C675B6155682729C9E08D51A0A7501105C \
+    8B44A05C308955D191956559A5CEE20A90348D47
+
+ENV HADOOP_VERSION 2.7.2
+ENV HADOOP_URL https://archive.apache.org/dist/hadoop/core/hadoop-$HADOOP_VERSION/hadoop-$HADOOP_VERSION.tar.gz
+
+RUN set -x \
+    && curl -fSL "$HADOOP_URL" -o /tmp/hadoop.tar.gz \
+    && curl -fSL "$HADOOP_URL.asc" -o /tmp/hadoop.tar.gz.asc \
+    && gpg --verify /tmp/hadoop.tar.gz.asc \
```
abhishekdas99: Do we need .asc file and verify that with gpg ?

Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+    && tar -xvf /tmp/hadoop.tar.gz -C /usr/lib/ \
+    && rm /tmp/hadoop.tar.gz*
+
+ENV HADOOP_PREFIX /usr/lib/hadoop
+
+RUN cd /usr/lib && mv hadoop-$HADOOP_VERSION hadoop2 && ln -s ./hadoop2 hadoop
+
+RUN mkdir $HADOOP_PREFIX/input
+RUN cp $HADOOP_PREFIX/etc/hadoop/*.xml $HADOOP_PREFIX/input
```
abhishekdas99: Remove this. Don't think this is needed

Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+
+ENV HADOOP_CLASSPATH=$HADOOP_CLASSPATH:/usr/lib/hadoop2/share/hadoop/tools/lib/*
+
+ENV HADOOP_CONF_DIR=/etc/hadoop
```
abhishekdas99: Remove this if unnecessary

Dockerfile
kvankayala: This installation can be done without gpg verification check. It is just a safety check.
I have update the file without gpg check

Dockerfile
kvankayala: Removed extra lines

Dockerfile
kvankayala: Removed extra lines

PR #182, opened 2018-07-20
Added implementation for PrestoClusterManagerV2

abhishekdas99:


abhishekdas99:
I have checked with @amoghmargoor .. Will also talk to @stagraqubole .. @amoghmargoor has raised some concerns for which we will create issues.

Issue #119, opened 2018-04-26
Add ability to disable staleness checks


wishnick:
Since our data is immutable we'd like to reduce the calls to S3 for files that are already cached to check if they have changed.

I took a small stab at trying to make this change and my best guess was that it would mean getting rid of [this call here](https://github.com/qubole/rubix/blob/4214e36265885d0b2d3ac6a98088c7ed7dbc6b85/rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java#L88) to the parent file system, which ultimately delegates down to [PrestoS3FileSystem here](https://github.com/prestodb/presto/blob/c12daee0cfd8580095c90eb4668e456eb304b5ae/presto-hive/src/main/java/com/facebook/presto/hive/s3/PrestoS3FileSystem.java#L292-L322) where it calls S3.

But I wasn't 100% sure if this was the right place or if there was a better place.

abhishekdas99:
We do make S3 calls (HEAD call) to get the filesize and lastmodified time. Lastmodified time is used for invalidation. In you case if the data is immutable, so we don't need to invalidate. But the [code you pointed](https://github.com/qubole/rubix/blob/4214e36265885d0b2d3ac6a98088c7ed7dbc6b85/rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java#L88) where we are making getFileStatus call, it also provides the file length which is needed when we are reading the file.

For eg, if the filesize is 1000 bytes and the client wants to read from offset 900 bytes and of length 200 bytes, then the file size comes into picture and limits the read till 1000 bytes.
So whenever we open an input stream, we get the above mentioned information from the remote call. And its not possible to completely get rid off the call.

One possible solution we can get is pass this remote call to bookkeeper server. In that case we can cache the file size as well and we can also add a flag to enable invalidation check. Caching the filesize and turning off file invalidation flag will not make getFileStatus call everytime when we open an input steam object.
In that case we need to add a new method in thrift layer.

cc @vrajat

stagraqubole:
There is a branch to cache file status and listings in rubix: https://github.com/stagraqubole/rubix/tree/cacheMD

That could also help this use case by only reading this information (once per node in worst case) and caching it for subsequent reads. We can work to bring in that change if it helps @wishnick @abhishekdas99

PR #183, opened 2018-07-24
new: usr: #119: Add ability to disable staleness check for files

abhishekdas99:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -444,6 +477,21 @@ public void onRemoval(final RemovalNotification<String, FileMetadata> notificati
...
+  private static void initializeFileInfoCache(final Configuration conf)
+  {
+    fileInfoCache = CacheBuilder.newBuilder()
+        .expireAfterWrite(36000, TimeUnit.SECONDS)
+        .removalListener(new RemovalListener<String, FileInfo>()
+        {
+          @Override
+          public void onRemoval(RemovalNotification<String, FileInfo> notification)
+          {
+            log.info("removed FileInfo for path " + notification.getKey() + " due to " + notification.getCause());
```
vrajat: Just want to make sure you mean to log every eviction at info level.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -319,6 +322,34 @@ public boolean readData(String remotePath, long offset, int length, long fileSiz
...
+        FileInfo info = new FileInfo(status.getLen(), status.getModificationTime());
+        return info;
+      }
+      catch (Exception e) {
+        log.error(String.format("Could not fetch FileStatus from remote file system for %s : %s", remotePath, Throwables.getStackTraceAsString(e)));
+      }
+    }
+    else {
+      try {
+        return fileInfoCache.get(remotePath, new FetchFileInfoCallable(remotePath, conf));
```
vrajat: Curious why you chose the Callable paradigm instead of the CacheLoader/Load fn. paradigm. Is it because of the conf variable ? I suggest you comment why you have chosen a different paradigm to the rest of the caches in this project.

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeper.java
```diff
@@ -138,4 +142,52 @@ public void testGetPrestoClusterManagerInValidInstance() throws Exception
...
+    FileInfo info = bookKeeper.getFileInfo(backendFilePath.toString());
+
+    assertTrue(info.getFileSize() == expectedFileSize, "FileSize was not equal to the expected value." +
+        " Got FileSize: " + info.getFileSize() + " Expected Value : " + expectedFileSize);
+
+    //Rewrite the file with half the data
+    DataGen.populateFile(backendFilePath.toString(), 2);
+
+    info = bookKeeper.getFileInfo(backendFilePath.toString());
+    assertTrue(info.getFileSize() == expectedFileSize, "FileSize was not equal to the expected value." +
```
vrajat: Is this proving that the file size is still 1 in the metadata ?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
abhishekdas99: I chose to implement through Callable because the other cache (FileMetadtaCache) was using the same. Read when to use Callable and CacheLoader/Load fn. Changed the implementation to CacheLoader/Load fn.

+1 for the review comment.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
abhishekdas99: Yes. Just logging when the FileInfo is expiring. We can move it to Debug level as well.

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeper.java
abhishekdas99: This is to show that although we have changed the file in the backend (which is half the original size), the fileinfo still has the old value. (Data is being cached)

abhishekdas99:
Added test on cache eviction through ttl.

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeper.java
```diff
@@ -138,4 +142,85 @@ public void testGetPrestoClusterManagerInValidInstance() throws Exception
...
+        " Got FileSize: " + info.getFileSize() + " Expected Value : " + expectedFileSize);
+
+    // Sleep for 5 sec to expire the cache
+    Thread.sleep(5000);
```
vrajat: You should use ticks here. Guava cache has a way to do that. There must be an example of that written by @jordanw-bq

jordanw-bq: @abhishekdas99 https://github.com/qubole/rubix/blob/6ebb35c2d9c44a48f2643f296a620fc9ea21bd85/rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestCoordinatorBookKeeper.java#L100

This test uses a `FakeTicker` to advance the cache in order to expire the entries. I had to implement a Mock(Coordinator)BookKeeper in order to supply a custom ticker, though (L123 of the same test class)

Issue #172, opened 2018-07-10
Add metrics for local data validation success/failure


dannykBQ:
**Requirements**

Validator to check for files and their accompanying metadata files
Metric for validation success & failure rate
Alert if validation success rate falls below certain threshold [CONFIRM THRESHOLD]

PR #190, opened 2018-08-01
 new: dev: #172: add service to validate cache files

jordanw-bq:


rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -80,6 +83,9 @@
   // default values
   private static final int DEFAULT_BLOCK_SIZE = 1 * 1024 * 1024; // 1MB
   private static final int DEFAULT_CLIENT_TIMEOUT = 10000; // ms
+  private static final String DEFAULT_CACHE_METADATA_FILE_SUFFIX = "_mdfile";
+  private static final int DEFAULT_CACHE_VALIDATION_INITIAL_DELAY = 30000; // ms
```
abhishekdas99: Is cache validation every 30sec very frequent ? Can we something like 5 min or 10 min ? @vrajat

jordanw-bq: @abhishekdas99 @vrajat My bad here. I think we had discussed 30min, but my mind was still stuck on seconds from the other intervals.

Updated the interval & initial delay to 30min

vrajat:
I need to come back to this PR a couple of more times. Looks good in first pass though.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/validation/CacheValidator.java
```diff
@@ -0,0 +1,108 @@
...
+    final ValidationResult allDisksResult = new ValidationResult();
+    for (int diskIndex = 0; diskIndex < maxDisks; diskIndex++) {
+      ValidatorFileVisitor validatorVisitor = new ValidatorFileVisitor(conf);
+      Files.walkFileTree(Paths.get(CacheUtil.getDirPath(diskIndex, conf)), validatorVisitor);
+
+      allDisksResult.addResult(validatorVisitor.getResult());
+    }
+
+    if (allDisksResult.getFailureCount() > 0) {
+      log.error("Validation Error!");
```
vrajat: This is not attached to Rubix health check and it will be done in a subsequent phase ?

jordanw-bq: This is now tied in with a health check created as part of the other validation PR. After reviewing with Das, we determined it would be easier to combine the two PRs to implement the health check logic. 
Apologies for not communicating this earlier.

jordanw-bq:
The metrics changes are only a name change; I felt the "METRIC_BOOKKEEPER" prefix for the enum names was making them too verbose

jordanw-bq:
Closing PR, since functionality was merged into #199

Issue #188, opened 2018-07-29
Initialize ClusterManager only from Coordinator BookKeeper module


abhishekdas99:
Currently ClusterManager is used to get a node index for a particular key using Consistent Hashing logic. This node index is used to get the actual node hostname from a list of nodes (which is also provided by ClusterManager). This hostname is used to figure out whether a file belongs to that particular node or not and also used to in providing hint to the scheduler to have data locality for tasks.

ClusterManager from each node invokes the respective apis to get the list of nodes and maintains the state of those nodes. This current architecture is not going to helpful in case where the ClusterManager needs to maintain a state (let say all nodes it has seen and current state of those nodes). In that case, we need to persist the info in local disk so that if the rubix daemon goes down, we don't lose its state.

So we are proposing an architecture where the master bookkeeper daemon will invoke the corresponding apis to get the list of nodes in the cluster and cache it for configurable time. All the worker bookkeepers are going to fetch it from the master bookkeeper instead of maintaining its own state. They can also cache this data for some time to reduce the number of calls to master bookkeeper.

PR #194, opened 2018-08-07
fix: dev: #188: Initialize ClusterManager only from BookKeeper module

abhishekdas99:
The main purpose of this PR is:
1. ClusterManagers will only get initialized from BookKeeper classes, so that we have it centralized.
2. The clustermanager has different responsibilities
a) Returns the nodes in the cluster (CoordinatorBookkeeper gets this info from the url mentioned in the cluster manager. WorkerBookkeeper gets this info from the coordinator and caches it in local. The main reason for this is to maintain the state of what nodes are live in the cluster now and have centralize place to take care of it.)
b) To provide the nodeindex for a particular key using consistent hashing. This needs a key and number of items in the consistent hashing ring. This logic is be invoked by both coordinator as well as worker..


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -29,19 +35,54 @@
...
     startHeartbeatService(conf);
+    initializeNodesCache(conf, ticker);
+  }
+
+  private void initializeNodesCache(final Configuration conf, final Ticker ticker)
+  {
+    ExecutorService executor = Executors.newSingleThreadExecutor();
+    nodeListCache = CacheBuilder.newBuilder()
+        .ticker(ticker)
+        .expireAfterWrite(5, TimeUnit.MINUTES)
```
abhishekdas99: This is hardcoded now. I will make this configurable.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -50,22 +92,89 @@ public void handleHeartbeat(String workerHostname)
...
+
+    try {
+      List<String> nodes = nodeListCache.get(clusterType);
+      int nodeIndex = clusterManager.getNodeIndex(nodes.size(), remotePath);
+      String hostName = nodes.get(nodeIndex);
+
+      return hostName;
+    }
+    catch (ExecutionException e) {
+      e.printStackTrace();
```
jordanw-bq: Default `catch` block; could use better logging

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestWorkerBookKeeper.java
```diff
@@ -96,6 +107,64 @@ public void testHandleHeartbeat_shouldNotBeHandled() throws FileNotFoundExceptio
...
+    mockList.add("changed_localhost");
+
+    doReturn(mockList).when(spyCoordinator).getNodeHostNames(anyInt());
+    hostName = workerBookKeeper.getClusterNodeHostName("remotepath", ClusterType.TEST_CLUSTER_MANAGER.ordinal());
+
+    assertTrue(hostName.equals(testLocalhost), "HostName is not correct from the cache");
+    ticker.advance(200, TimeUnit.SECONDS);
+
+    hostName = workerBookKeeper.getClusterNodeHostName("remotepath", ClusterType.TEST_CLUSTER_MANAGER.ordinal());
+    assertTrue(hostName.equals("changed_localhost"), "HostName is not refreshed from Coordinator");
```
jordanw-bq: This string should be extracted for consistency (like `testLocalhost`)

abhishekdas99:
@stagraqubole @amoghmargoor Can we review this one ?

abhishekdas99:
Ping again for review.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
```diff
@@ -84,6 +87,43 @@ public void handleHeartbeat(String workerHostname, HeartbeatStatus heartbeatStat
     }
   }
 
+  @Override
+  public List<String> getNodeHostNames(int clusterType)
```
stagraqubole: Why do we need this in interface? This could be a private method inside CoordBookKeeper

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
stagraqubole: +1, log rather then just print

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -44,6 +92,38 @@ public void handleHeartbeat(String workerHostname, HeartbeatStatus request)
...
+    }
+
+    return null;
```
stagraqubole: We can potentially return null from here but the consumer getCacheStatus() blindly puts the return value in map which will throw a NPE

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -52,4 +132,41 @@ private void startHeartbeatService(Configuration conf, MetricRegistry metrics, B
...
+   * @param bookKeeperFactory   The factory to use for creating a BookKeeper client.
+   * @return The client used for communication with the master node.
+   */
+  private static RetryingBookkeeperClient initializeClientWithRetry(BookKeeperFactory bookKeeperFactory, Configuration conf,
+                                                             String hostName)
+  {
+    final int retryInterval = CacheConfig.getServiceRetryInterval(conf);
+    final int maxRetries = CacheConfig.getServiceMaxRetries(conf);
+
+    for (int failedStarts = 0; failedStarts < maxRetries; ) {
```
stagraqubole: why do you need to do all the retries here? RetryingBookkeeperClient is meant to do it.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
abhishekdas99: We need this because worker bookkeeper invokes this method to populate its local cache.
https://github.com/qubole/rubix/pull/194/files#diff-85292be7f2f40e0c26f9f055cbadbcb0R79

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
abhishekdas99: Done

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
abhishekdas99: Good point. It is handled in CahingFileSystem.getFileBlockLocations but not in BookKeeper.getCacheStatus. Introduced a new block location called Unknown. If the hostname is null, we just read data from remote object store

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
abhishekdas99: The retries are needed because even with the retries in RetryingBookkeeperClient, the worker might not be able to connect to master. Reason being master bookkeeper might not be up whereas the worker bookkeeper is up. This call is only made during the start of the worker bookkeeper.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
abhishekdas99: The logic in getClusterNodeHostName is 
1. get the list of nodes.
2. invoke cluster manager getNodeIndex with number of nodes and the file path to get the nodeIndex
3. Use nodeIndex to get the actual hostname from the list of nodes.

The difference between Coordinator and Worker implementation is at step 1 (how the list of nodes is fetched)
We can extract step 2 and 3  to a common method. Thats only two lines of code. 
The main logic of calculating the node index and getting the list of nodes are already pushed into cluster manager classes.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
stagraqubole: In that case we need to push the transport creation and opening into RetryingBookkeeperClient from BookKeeperFactory.createBookKeeperClient and extend RetryingBookkeeperClient retry logic to also take sleep interval.

abhishekdas99: Currently the retry logic in RetryingBookkeeperClient is only during the method invoke. We create the transport and open it without any retry. So createBookKeeperClient doesn't have any retry logic.
I agree with you that all the retry logic should be in one place and that retry should take sleep interval to make it work in all the cases. But I think that is outside the scope of this PR. So in my opinion, we should open a new issue to address this problem.

abhishekdas99: Updated 

> In that case we need to push the transport creation and opening into RetryingBookkeeperClient from BookKeeperFactory.createBookKeeperClient and extend RetryingBookkeeperClient retry logic to also take sleep interval.

Done

abhishekdas99: @jordanw-bq I have changed the retrying logic of creating the bookkeeper client. Let me know if you have any comments

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -14,28 +14,80 @@
...
+    initializeNodesCache(conf, ticker);
+  }
+
+  private void initializeNodesCache(final Configuration conf, final Ticker ticker)
+  {
+    int expiryPeriod = CacheConfig.getWorkerNodeInfoExpiryPeriod(conf);
+    ExecutorService executor = Executors.newSingleThreadExecutor();
+    nodeListCache = CacheBuilder.newBuilder()
+        .ticker(ticker)
+        .expireAfterWrite(expiryPeriod, TimeUnit.SECONDS)
```
stagraqubole: refreshAfterWrite would suit better than expireAfterWrite. Node additions and removals are less frequent and we would pay a penalty of reloading the value every 5min, async refresh with a fallback to direct path once in a while would be okay

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -306,22 +248,36 @@ else if (start >= 0L && len >= 0L) {
...
+
+          RetryingBookkeeperClient client = new BookKeeperFactory().createBookKeeperClient(conf);
+
+          for (long i = 0; i < file.getLen(); i = i + splitSize) {
+            long end = i + splitSize;
+            if (end > file.getLen()) {
+              end = file.getLen();
+            }
+            String key = file.getPath().toString() + i + end;
+            String hostName = client.getClusterNodeHostName(key, clusterType.ordinal());
```
stagraqubole: any impact on query performance after this change? We have added more network calls in critical sections after this change.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
abhishekdas99: +1

abhishekdas99: Done.

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
abhishekdas99: We haven't measured the performance yet but this is not over the network (cross node every time). Most of the time getFileBlockLocations gets invoked in the coordinator and in that case its a call to bookkeeper daemon running on the same node. For the case where this code is running in the Worker node, it will be a call to the coordinator only when the cache has expired.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -118,18 +104,31 @@ public boolean isBookKeeperAlive()
...
     initializeCache(conf, ticker);
-    cleanupOldCacheFiles(conf);
     fetchProcessor = new RemoteFetchProcessor(this, metrics, conf);
     fetchProcessor.startAsync();
   }
 
+  private void setupCacheDirectory(Configuration conf) throws BookKeeperInitializationException
+  {
+    try {
+      CacheUtil.createCacheDirectories(conf);
```
kvankayala: With this change, we can remove creating the directories from bookkeeper startup script.

jordanw-bq: This change is a reordering and does not change any functionality (was previously found in `initializeCache()`).
We still need the parent directories created through the BK startup script in order to create the cache directories.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
```diff
@@ -43,24 +52,91 @@
...
+
+  private void initializeCoordinator() throws CoordinatorInitializationException
+  {
+    initializeClusterManager();
+    setCurrentNodeName();
     registerMetrics();
   }
 
+  @VisibleForTesting
+  protected void initializeClusterManager() throws CoordinatorInitializationException
```
kvankayala: This function is used in initializeCoordinator (line:79) which is not a test function. So, I think you can remove **VisibleForTesting** annotation

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
```diff
@@ -43,24 +52,91 @@
...
+      nodeName = InetAddress.getLocalHost().getHostAddress();
+    }
+    catch (UnknownHostException e) {
+      log.warn("Could not get Host Address", e);
+      nodeName = ClusterUtil.getMasterHostname(conf);
+    }
+  }
+
+  @VisibleForTesting
+  protected ClusterManager getClusterManagerInstance(ClusterType clusterType, Configuration config)
```
kvankayala: Same here for **VisibleForTesting**

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -14,28 +14,131 @@
...
+            if (client == null) {
+              client = bookKeeperFactory.createBookKeeperClient(masterHostname, conf);
+              threadLocalConnections.set(client);
+            }
+            return client.getOwnerNodeForPath(s);
+          }
+        });
+  }
+
+  public WorkerBookKeeper(Configuration conf, MetricRegistry metrics, BookKeeperFactory factory) throws BookKeeperInitializationException
```
jordanw-bq: Constructors should be at the top of the class

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
abhishekdas99: The `@VisibleForTesting` annotation "indicates that the visibility of a type or member has been relaxed to make the code testable."  
`initializeCoordinator` is used in test classes and making it protected should work. `@VisibleForTesting` tag is just to make sure we document why the scope of this method is relaxed.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
```diff
@@ -84,6 +162,20 @@ public void handleHeartbeat(String workerHostname, HeartbeatStatus heartbeatStat
...
+  @Override
+  public Map<String, NodeState> getClusterNodes()
+  {
+    return getClusterManager().getNodes();
+  }
+
+  @Override
+  public String getOwnerNodeForPath(String remotePathKey)
+  {
+    Map<String, NodeState> nodesMap = getClusterNodes();
```
stagraqubole: 1. Move getOwnerNodeForPath to BookKeeper.java
2. Make getClusterNodes abstract method in BookKeeper.java
3. Remove getOwnerNodeForPath from WorkerBookKeeper and CoordinatorBookKeeper

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -211,24 +210,6 @@ public Long getValue()
   @Override
   public List<BlockLocation> getCacheStatus(CacheStatusRequest request) throws TException
```
stagraqubole: We are no more using cluster_type from CacheStatusRequest, so the thrift idl and the users of CacheStatusRequest should be changed to not fill it.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
```diff
@@ -43,24 +54,91 @@
...
   @VisibleForTesting
-  public CoordinatorBookKeeper(Configuration conf, MetricRegistry metrics, Ticker ticker) throws FileNotFoundException
+  protected CoordinatorBookKeeper(Configuration conf, MetricRegistry metrics, Ticker ticker) throws BookKeeperInitializationException
   {
     super(conf, metrics, ticker);
     this.isValidationEnabled = CacheConfig.isValidationEnabled(conf);
     this.liveWorkerCache = createHealthCache(conf, ticker);
     this.cachingValidatedWorkerCache = createHealthCache(conf, ticker);
     this.fileValidatedWorkerCache = createHealthCache(conf, ticker);
+    this.clusterType = CacheConfig.getClusterType(conf);
```
stagraqubole: default cluster type in configs is TEST_CLUSTER_MANAGER ... This change will need a change in all deployments of rubix to set the configs right, any scripts setting up rubix will need to be changed.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -14,28 +14,138 @@
...
+  private LoadingCache<String, Map<String, NodeState>> nodeStateMap;
   private HeartbeatService heartbeatService;
+  private BookKeeperFactory bookKeeperFactory;
+  private RetryingBookkeeperClient coordinatorClient;
+  private ThreadLocal<RetryingBookkeeperClient> threadLocalConnections;
+  // The hostname of the master node.
+  private String masterHostname;
+  private int clusterType;
+  String nodeHostName;
+  String nodeHostAddress;
```
stagraqubole: We can remove nodeHostAddress from class variable

stagraqubole: nodeHostName is also not needed here, nodeName is present in BookKeeper and is the one at class level

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -44,12 +154,58 @@ public void handleHeartbeat(String workerHostname, HeartbeatStatus request)
     throw new UnsupportedOperationException("Worker node should not handle heartbeat");
   }
 
+  @Override
+  public Map<String, NodeState> getClusterNodes()
+  {
+    RetryingBookkeeperClient client = threadLocalConnections.get();
+    try {
+      if (client == null) {
+        client = createBookKeeperClientWithRetry(bookKeeperFactory, masterHostname, conf);
```
stagraqubole: you are doing this already in getter of nodeStateMap cache, done need to check and set client here

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -14,28 +14,138 @@
...
+import static com.qubole.rubix.spi.ClusterType.TEST_CLUSTER_MANAGER_MULTINODE;
 
 public class WorkerBookKeeper extends BookKeeper
 {
   private static Log log = LogFactory.getLog(WorkerBookKeeper.class);
+  private LoadingCache<String, Map<String, NodeState>> nodeStateMap;
   private HeartbeatService heartbeatService;
+  private BookKeeperFactory bookKeeperFactory;
+  private RetryingBookkeeperClient coordinatorClient;
+  private ThreadLocal<RetryingBookkeeperClient> threadLocalConnections;
```
stagraqubole: ThreadLocals are nasty and can cause leaks. We are also causing leaks here, not closing BookKeeperClient

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
abhishekdas99: In a separate PR, I am removing all cluster_type reference where its not required.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
abhishekdas99: The script that starts the rubix daemons will pass the cluster type. So for unit tests, it will be set to TEST_CLUSTER_MANAGER, this will be overridden with the incoming config

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
abhishekdas99: Made these variables local to `setCurrentNodeName`

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -211,24 +212,6 @@ public Long getValue()
...
-      return null;
-    }
-
-    if (currentNodeIndex == -1 || nodes == null) {
-      log.error("Initialization not done");
-      return null;
-    }
-
     Map<Long, String> blockSplits = new HashMap<>();
     long blockNumber = 0;
```
stagraqubole: incrMetrics is unused

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/utils/ConsistentHashUtil.java
```diff
@@ -0,0 +1,94 @@
...
+  private static Log log = LogFactory.getLog(ConsistentHashUtil.class.getName());
+
+  private ConsistentHashUtil()
+  {
+    //
+  }
+
+  public static String getHashedNodeForKey(Map<String, NodeState> nodesMap, String key)
+  {
+    List<String> nodes = Lists.newArrayList(nodesMap.keySet().toArray(new String[0]));
```
stagraqubole: 1. We are creating too many objects, new ArrayList here and another one in getNodeIndex. And this is per getCacheStatus request
2. nodeMap.keySet() has no ordering, shouldnt we be ordering the list of nodes before using them?
3. Better than sorting and creating new lists so many times, we can make BookKeeper.getClusterNodes() return a List<Node> where Node class has nodeName and nodeState. And then we can push the logic in clusterManager to always output a ordered List<Node>.

stagraqubole: ClusterManager.getNodes used to return sorted list earlier (there is a comment as well)

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
stagraqubole: This is still not handled ... Adding 'if (blockSplits.get(split) == null) ` in BookKeeper.getCacheStatus will not be enough because you are still doing this there:
`      String hostName = getOwnerNodeForPath(key);
      blockSplits.put(blockNumber, hostName);
`

And Map.put(T, null) will throw NullPointerException

rubix-spi/src/main/java/com/qubole/rubix/spi/ClusterManager.java
```diff
@@ -31,53 +29,26 @@
...
   public int getNodeRefreshTime()
   {
     return nodeRefreshTime;
   }
 
-  public abstract boolean isMaster()
-      throws ExecutionException;
-
   // Nodes format as per the note above
   // Should return sorted list
```
jordanw-bq: What note does this comment refer to? Seems like this could use rephrasing or removal

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
abhishekdas99: `testGetCacheStatusWhenHostNameIsNull` proves we can put null as value in the map.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/utils/ConsistentHashUtil.java
abhishekdas99: Changed the implementation to use List

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -138,11 +152,13 @@ public void stopServer()
       log.error("Metrics reporters could not be closed", e);
     }
     server.stop();
+    log.debug("Bookkeeper Server Stopped");
```
jordanw-bq: Suggestion: Mirror the phrasing of the start log line
```suggestion
    log.debug("Stopped BookKeeperServer");
```

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -14,29 +14,131 @@
...
+  {
+    this(conf, bookKeeperMetrics, Ticker.systemTicker(), factory);
+  }
 
-  public WorkerBookKeeper(Configuration conf, BookKeeperMetrics bookKeeperMetrics) throws FileNotFoundException
+  public WorkerBookKeeper(Configuration conf, BookKeeperMetrics bookKeeperMetrics, Ticker ticker, BookKeeperFactory factory)
+      throws BookKeeperInitializationException
   {
-    this(conf, bookKeeperMetrics, new BookKeeperFactory());
+    super(conf, bookKeeperMetrics, Ticker.systemTicker());
```
jordanw-bq: `WorkerBookKeeper`s will always get a System ticker for their file metadata & file info caches. Is this the intended behaviour?

abhishekdas99: Good catch. Fixed

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -14,29 +14,131 @@
...
+import static com.qubole.rubix.spi.ClusterType.TEST_CLUSTER_MANAGER;
+import static com.qubole.rubix.spi.ClusterType.TEST_CLUSTER_MANAGER_MULTINODE;
 
 public class WorkerBookKeeper extends BookKeeper
 {
   private static Log log = LogFactory.getLog(WorkerBookKeeper.class);
+  private LoadingCache<String, List<ClusterNode>> nodeStateMap;
   private HeartbeatService heartbeatService;
+  private BookKeeperFactory bookKeeperFactory;
+  private RetryingBookkeeperClient coordinatorClient;
```
jordanw-bq: `coordinatorClient` is unused

rubix-presto/src/main/java/com/qubole/rubix/presto/PrestoClusterManager.java
```diff
@@ -175,18 +165,17 @@ public void initialize(Configuration conf)
...
               if (hosts.isEmpty()) {
                 // case of master only cluster
-                hosts.add(InetAddress.getLocalHost().getHostAddress());
+                hosts.add(new ClusterNode(InetAddress.getLocalHost().getHostAddress(), NodeState.ACTIVE));
               }
-              List<String> hostList = Lists.newArrayList(hosts.toArray(new String[0]));
-              Collections.sort(hostList);
-              return hostList;
+
+              return hosts;
```
stagraqubole: We are not sorting the list anymore and neither do we do it when when use it in bookkeeper. Wouldn't this cause inconsistency because node indexes are not guaranteed to be same for the same membership of cluster.

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -306,22 +252,36 @@ else if (start >= 0L && len >= 0L) {
...
+
+          RetryingBookkeeperClient client = new BookKeeperFactory().createBookKeeperClient(conf);
+
+          for (long i = 0; i < file.getLen(); i = i + splitSize) {
+            long end = i + splitSize;
+            if (end > file.getLen()) {
+              end = file.getLen();
+            }
+            String key = file.getPath().toString() + i + end;
+            String hostName = client.getOwnerNodeForPath(key);
```
stagraqubole: Is there a plan to benchmark this change? Thrift calls in planning phase are bit of a concern given that we see high socket read times once in a while in read phase.

rubix-spi/src/main/thrift/bookkeeper.thrift
```diff
@@ -6,7 +6,18 @@ typedef i32 int
 enum Location {
 CACHED,
 LOCAL,
-NON_LOCAL
+NON_LOCAL,
+UNKNOWN
+}
+
+enum NodeState {
```
jordanw-bq: We should start standardizing the indenting for code blocks in this file (either 4 spaces to match the rest of this file, or 2 spaces to match our Java code style)

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -87,6 +87,10 @@
   private static final String KEY_MAX_CACHE_SIZE = "rubix.cache.max.size";
   private static final String KEY_CACHE_FILE_SPLIT_SIZE = "rubix.cache.filesplit.size";
   private static final String KEY_CLUSTER_NODE_REFRESH_TIME = "rubix.cluster.node.refresh.time";
+  private static final String KEY_WORKER_NODEINFO_EXPIRY_PERIOD = "rubix.worker.nodeinfo.expiry.period";
+  private static final String KEY_RUBIX_CLUSTER_TYPE = "rubix.cluster.type";
+  private static final String KEY_RUBIX_WORKER_NODELIST_FETCH_MAX_RETRIES = "rubix.worker.nodelist.fetch.max-retries";
+  private static final String KEY_RUBIX_WORKER_NODELIST_FETCH_WAIT_INTERVAL = "rubix.worker.nodelist.fetch.wait.interval";
```
jordanw-bq: Add documentation for configuration options where appropriate (especially for `rubix.cluster.type`, since this option takes an integer instead of a meaningful name)

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
```diff
@@ -42,24 +52,91 @@
   protected Cache<String, Boolean> fileValidatedWorkerCache;
 
   private final boolean isValidationEnabled;
+  private static Integer lock = 1;
+  protected ClusterManager clusterManager;
+  int clusterType;
```
jordanw-bq: Decrease the visibility of this field

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -14,29 +14,166 @@
...
+  // The hostname of the master node.
+  private String masterHostname;
+  private int clusterType;
+  RetryingBookkeeperClient client;
```
jordanw-bq: Decrease visibility for this field.

PR #197, opened 2018-08-22
 fix: usr: #196: Evict files from local when the cache is full

abhishekdas99:


rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -78,6 +78,7 @@
   private static final String KEY_DUMMY_CLUSTER_MANAGER = "rubix.dummy.clustermanager.class";
   private static final String KEY_ENABLE_FILE_STALESSNESS_CHECK = "rubix.enable.file.staleness-check";
   private static final String KEY_STALE_FILEINFO_EXPIRY_PERIOD = "rubix.stale.fileinfo.expiry.period";
+  private static final String KEY_CLEANUP_FILES_DURING_START = "rubix.cleanup.files.duing.start";
```
kvankayala: Typo! I think the configuration you want to name is "rubix.cleanup.files.**during**.start" instead of "rubix.cleanup.files.**duing**.start"

vrajat:
I did one pass. I dont have any feedback. I am still getting used to this part of the code. Tests will help. I know unit tests will help. @kvankayala is it possible to write integration tests now ? This PR is a good candidate for integrations tests if the framework is ready.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -113,10 +116,36 @@ public BookKeeper(Configuration conf, MetricRegistry metrics) throws FileNotFoun
...
+    if (CacheConfig.isCleanupFilesDuringStartEnabled(conf)) {
+      try {
+        int numDisks = CacheConfig.getCacheMaxDisks(conf);
+        String dirSuffix = CacheConfig.getCacheDataDirSuffix(conf);
+        List<String> dirPrefixList = CacheUtil.getDirPrefixList(conf);
+
+        for (String dirPrefix : dirPrefixList) {
+          for (int i = 0; i < numDisks; i++) {
+            java.nio.file.Path path = Paths.get(dirPrefix + i, dirSuffix, "*");
+            Files.walkFileTree(path, new DeleteFileVisitor());
```
stagraqubole: Wouldnt it be faster to just to `rm -rf` on all parent directories under fcache? This could get very slow in some cases.

abhishekdas99: Done

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -585,27 +617,37 @@ public void onRemoval(RemovalNotification<String, FileMetadata> notification)
...
+  // deleteCachedFile determines whether to delete the actual file from the local filesystem or not
+  private static void invalidateFileMetadata(String key, boolean deleteCachedFile)
   {
     // We might come in here with cache not initialized e.g. fs.create
     if (fileMetadataCache != null) {
-      fileMetadataCache.invalidate(p);
+      FileMetadata metadata = fileMetadataCache.getIfPresent(key);
+      fileMetadataCache.invalidate(key);
+
+      if (metadata != null && deleteCachedFile) {
```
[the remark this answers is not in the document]
abhishekdas99: I have put the same key and a new object of FileMetadata (that will take care of the weight). This will cause the older value to be REPLACEd. We will get Removal Notification with cause REPLACE at that time. 

So Invalidate is going to evict the file with cause EXPLICIT  and we will delete the files if the cause is not REPLACE

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -325,14 +354,16 @@ public void setAllCached(String remotePath, long fileLength, long lastModified,
       return;
     }
     if (isInvalidationRequired(md.getLastModified(), lastModified)) {
-      invalidate(remotePath);
+      invalidateFileMetadata(remotePath);
```
stagraqubole: With new code, this will not delete the file now due to EXPLICIT eviction while we need the file to be deleted due to change in data

stagraqubole: nvm, I see it handled

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -570,7 +596,19 @@ public void onRemoval(RemovalNotification<String, FileMetadata> notification)
       FileMetadata md = notification.getValue();
       try {
         md.closeAndCleanup(notification.getCause(), fileMetadataCache);
-        cacheEvictionCount.inc();
+        switch (notification.getCause()) {
+          case EXPLICIT:
+            cacheInvalidationCount.inc();
+            break;
+          case SIZE:
+            cacheEvictionCount.inc();
```
stagraqubole: We should differentiate SIZE from EXPIRED, users would be more interested in eviction count due to SIZE

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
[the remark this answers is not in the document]
stagraqubole: 👍

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
stagraqubole: But user can change the default. If you plan to get rid of count in case of EXPIRED case then this is fine but I would then suggest to rename the metric a bit to show it is for SIZE based expiry.

stagraqubole: Dont need to do it now if it needs more changes than just rename, in that case can take it up separately

abhishekdas99:
@vrajat I am writing the unit tests for this. Also we are working on the integration test framework and we will take this test up first when it is ready

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -192,10 +221,10 @@ public Integer getValue()
 
     FileMetadata md;
     try {
-      md = fileMetadataCache.get(remotePath, new CreateFileMetadataCallable(remotePath, fileLength, lastModified, conf));
+      md = fileMetadataCache.get(remotePath, new CreateFileMetadataCallable(remotePath, fileLength, lastModified, 0, conf));
```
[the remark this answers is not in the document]
stagraqubole: Not applicable

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
abhishekdas99: Done

Issue #173, opened 2018-07-10
Report metrics for verifying caching read behavior


dannykBQ:
* Adding another liveness metric for cache read. 
* Liveness: Read a block not cached, next read is cached and then purge. 
* Run duration: 30 mins

PR #199, opened 2018-09-05
 new: dev: #173: report metric for verifying caching behavior

jordanw-bq:
* Tested on Qubole Presto cluster

### DIFF
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
@@ -84,14 +84,13 @@ public abstract class BookKeeper implements BookKeeperService.Iface
   String nodeName;
   static String nodeHostName;
   static String nodeHostAddress;
-  private Configuration conf;
+  private final Configuration conf;
   private static Integer lock = 1;
   private List<String> nodes;
   int currentNodeIndex = -1;
   static long splitSize;
-  private RemoteFetchProcessor fetchProcessor;
-  private CachingBehaviorValidator cachingBehaviorValidator;
-  private Ticker ticker;
+  private final RemoteFetchProcessor fetchProcessor;
+  private final Ticker ticker;
 
   // Registry for gathering & storing necessary metrics
   protected final MetricRegistry metrics;
@@ -119,7 +118,15 @@ public abstract class BookKeeper implements BookKeeperService.Iface
     initializeMetrics();
     initializeCache(conf, ticker);
     cleanupOldCacheFiles(conf);
-    startServices();
+
+    fetchProcessor = new RemoteFetchProcessor(conf);
+    fetchProcessor.startAsync();
+
+    if (CacheConfig.isCachingBehaviorValidationEnabled(conf)) {
+      log.debug("Starting caching behavior validation");
+      CachingBehaviorValidator cachingBehaviorValidator = new CachingBehaviorValidator(conf, metrics, this);
+      cachingBehaviorValidator.startAsync();
+    }
   }
 
   // Cleanup the cached files that were downloaded as a part of previous bookkeeper session.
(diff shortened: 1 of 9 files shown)


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -239,23 +247,32 @@ public Integer getValue()
 
     //TODO: Store Node indices too, i.e. split to nodeIndex map and compare indices here instead of strings(nodenames).
 
+    final boolean isValidating = isValidatingCachingBehavior(remotePath);
```
vrajat: Why is this conditional required ?

jordanw-bq: Validation makes calls to BookKeeper methods, which affects metrics values. We need this in order to prevent these metrics from being incremented during validation; otherwise this would skew the metrics counts

vrajat: Got it. I cant think of a better solution than adding if conditions. My worry is that a future change to add metrics will forget to add the if condition. The only minor improvement I can think of is to pass a flag to not change metrics. instead of `isValidatingCachingBehaviour`

vrajat: Feel free to ignore that though. It will not make a massive change to the current approach.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
```diff
@@ -35,43 +36,101 @@
...
   @VisibleForTesting
-  CoordinatorBookKeeper(Configuration conf, MetricRegistry metrics, Ticker ticker) throws FileNotFoundException
+  public CoordinatorBookKeeper(Configuration conf, MetricRegistry metrics, Ticker ticker) throws FileNotFoundException
   {
     super(conf, metrics, ticker);
-    this.liveWorkerCache = CacheBuilder.newBuilder()
-        .ticker(ticker)
-        .expireAfterWrite(CacheConfig.getWorkerLivenessExpiry(conf), TimeUnit.MILLISECONDS)
-        .build();
+    this.liveWorkerCache = createHealthCache(conf, ticker);
```
abhishekdas99: Why we are invoking `createHealthCache` three times with the same arguments ?

jordanw-bq: Since the live worker and validation caches all share the same settings, this centralizes the creation logic.
I considered using just one cache for storing all the validation results, but this would require more processing to determine the number of validated workers when reporting metrics values.

rubix-common/src/main/java/com/qubole/rubix/common/metrics/BookKeeperMetrics.java
```diff
@@ -218,15 +218,17 @@ public String getMetricName()
...
   /**
-   * Enum for metrics relating to daemon & service liveness.
+   * Enum for metrics relating to daemon & service health.
    */
-  public enum LivenessMetric
+  public enum HealthMetric
   {
-    METRIC_BOOKKEEPER_LIVE_WORKER_GAUGE("rubix.bookkeeper.live_workers.gauge");
+    LIVE_WORKER_GAUGE("rubix.bookkeeper.live_workers.gauge"),
+    CACHING_VALIDATED_WORKER_GAUGE("rubix.bookkeeper.caching_validated_workers.gauge"),
```
kvankayala: Can you please rename the metric with new convention ?
Ex: Instead of rubix.bookkeeper.caching_validated_workers.**gauge** , it should be
rubix.bookkeeper.**gauge**.caching_validated_workers

rubix-common/src/main/java/com/qubole/rubix/common/metrics/BookKeeperMetrics.java
```diff
@@ -237,14 +239,49 @@ public String getMetricName()
...
+      return names;
+    }
+  }
+
+  /**
+   * Enum for metrics relating to validation.
+   */
+  public enum ValidationMetric
+  {
+    CACHING_VALIDATION_SUCCESS_GAUGE("rubix.bookkeeper.caching_validation_success.gauge"),
```
kvankayala: Here also, Can you please rename the metric with new convention ?

rubix-common/src/main/java/com/qubole/rubix/common/metrics/BookKeeperMetrics.java
```diff
@@ -237,14 +239,49 @@ public String getMetricName()
...
+    FILE_VALIDATION_SUCCESS_GAUGE("rubix.bookkeeper.file_validation_success.gauge");
```
kvankayala: here also, Can you please rename the metric with new convention ?

PR #202, opened 2018-09-15
fix: dev: ISS-201: Return total bytes read from FileDownloadRequest to update the cache metadata

abhishekdas99:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -117,7 +117,7 @@ public BookKeeper(Configuration conf, MetricRegistry metrics) throws FileNotFoun
     initializeMetrics();
     initializeCache(conf, ticker);
     cleanupOldCacheFiles(conf);
-    fetchProcessor = new RemoteFetchProcessor(conf);
+    fetchProcessor = new RemoteFetchProcessor(this, conf);
```
abhishekdas99: Main reason to pass the bookkeeper object here is to get rid off the thrift call from FileDownloadRequestChain. FileDownloadRequestChain has been moved to bookkeeper module. Previously we were calling setAllCached through thrift layer. With this change we already have a reference of bookkeeper so we can invoke bookkeeper.setAllCached to avoid the thrift call. 
Both these are running in the the same jvm. So avoid the call through the thrift layer

vrajat: Is this change orthogonal to the issue ? I am confused because the method `processDownloadRequests` doesnt seem to require this change. Also where will this be used ? I couldnt find a user of `processDownloadRequests` at least in this PR. Is that coming in a subsequent PR ?

abhishekdas99: yes this change (passing bookkeeper instance) is orthogonal to the issue being reported. The change needed for that is to return totalRequestedRead instead of 0.
But while writing unit test, figured this issue and hence fixing this.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloadRequestChain.java
```diff
@@ -102,22 +110,22 @@ public Integer call() throws IOException
...
         //log.info("Processing request of  " + readRequest.getBackendReadLength() + " from " + readRequest.backendReadStart);
         readBytes = copyIntoCache(inputStream, fileChannel, readRequest.getBackendReadLength(),
-            readRequest.backendReadStart);
+            readRequest.getBackendReadStart());
         totalRequestedRead += readBytes;
       }
       log.info("Downloaded " + totalRequestedRead + " bytes of file " + remotePath);
       log.debug("RemoteFetchRequest took : " + (System.currentTimeMillis() - startTime) + " msecs ");
-      return 0;
+      return totalRequestedRead;
```
abhishekdas99: This is the only fix that is needed for this issue

PR #207, opened 2018-09-22
fix: dev: #206: Remove ClusterManager dependency from CachingPrestoNativeAzureFileSystem

abhishekdas99:


rubix-presto/src/main/java/com/qubole/rubix/presto/CachingPrestoS3FileSystem.java
```diff
@@ -37,7 +37,7 @@ public CachingPrestoS3FileSystem()
   public void initialize(URI uri, Configuration conf) throws IOException
   {
     try {
-      initializeClusterManager(conf, ClusterType.HADOOP2_CLUSTER_MANAGER);
```
vrajat: I am surprised that this hasnt caused problems before.

abhishekdas99: This is a recent change and has not been released :)

abhishekdas99: Tried writing unit tests but its hard as we need to initialize the PrestoS3FileSystem

vrajat: Too much mocking ? I am wondering if you can check that the base class is consistent at least.

abhishekdas99: Tried mocking the class but the defeats the purpose checking scenarios like this. Have to figure out how to run these kind of check quickly.

Issue #209, opened 2018-09-27
LocalCacheRead metrics is redundantly incremented in BookKeeper daemon


abhishekdas99:
Consider this workflow:
client invokes getCacheStatus in bookkeeper daemon running on the same node (say N1) for a particular file.
if the file doesn't belong to the same node (N1), this will be a non-local read for that node. Metrics get incremented based on that (nonlocalread in this case)
the client invokes LocalDataTransferServer running on the different node (N2). 
LDTS running on N2 internally invokes getCacheStatus in bookkeeper running on N2
getCacheStatus again increments metrics (cachedread at this time)

In this scenario, the cachedread on N2 should not be incremented as this request has been counted as nonlocalread for N1

PR #210, opened 2018-09-27
fix: usr: #209: Increment cache request metrics based on a flag to remove redundant counting

abhishekdas99:


jordanw-bq:
Looks good to me

abhishekdas99:
@vrajat I agree with you. I am planning to revisit this code later to make the changes.

Issue #203, opened 2018-09-15
Implement metrics for Asynchronous cache warmup


abhishekdas99:
We need metrics such as total async request received, total async request processed, total downloaded data asynchronously, async queue size etc.

PR #211, opened 2018-10-03
fix: usr: #203: Add metrics for parallel cache warmup

abhishekdas99:


jordanw-bq:
Looks good to me

amoghmargoor:
LGTM ... just a minor nit. suggest using @VisbleForTesting for methods created for testing and make them private so that external clients don't use it.

Issue #204, opened 2018-09-15
Some task attempts are failing with ArrayIndexOutOfBoundsException with Asynchronous Cache Warmup enabled


abhishekdas99:
Some tasks of a job that is warming up the cache are failing with ArrayIndexOutOfBoundsException

**Job Type :** Hive on Tez Query (The same thing is reproducible with Spark as well)
**Initial Run or Subsequent Run:** Initial Run
**Status of the Job:** The job succeeds (with some task attempt failure)
**Status of subsequent run of the same job:** The subsequent runs of the same jobs succeed.
**Data Corruption:** No

The stack of the exception:

```
2018-09-06 20:32:40,938 INFO [TezChild] task.TezTaskRunner: Encounted an error while executing task: attempt_1536264376320_0001_1_02_000044_2
java.lang.RuntimeException: java.lang.RuntimeException: java.io.IOException: java.lang.ArrayIndexOutOfBoundsException
	at org.apache.hadoop.hive.ql.exec.tez.TezProcessor.initializeAndRunProcessor(TezProcessor.java:171)
	at org.apache.hadoop.hive.ql.exec.tez.TezProcessor.run(TezProcessor.java:137)
	at org.apache.tez.runtime.LogicalIOProcessorRuntimeTask.run(LogicalIOProcessorRuntimeTask.java:337)
	at org.apache.tez.runtime.task.TezTaskRunner$TaskRunnerCallable$1.run(TezTaskRunner.java:179)
	at org.apache.tez.runtime.task.TezTaskRunner$TaskRunnerCallable$1.run(TezTaskRunner.java:171)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:421)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1635)
	at org.apache.tez.runtime.task.TezTaskRunner$TaskRunnerCallable.callInternal(TezTaskRunner.java:171)
	at org.apache.tez.runtime.task.TezTaskRunner$TaskRunnerCallable.callInternal(TezTaskRunner.java:167)
	at org.apache.tez.common.CallableWithNdc.call(CallableWithNdc.java:36)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1152)
[...]

PR #212, opened 2018-10-03
fix: dev: #204: Clone read request if it is shared across read request chains

abhishekdas99:


jordanw-bq:
LGTM

vrajat:
Is it possible to add a reproducer ? Or too hard to replicate ?

abhishekdas99:
> Is it possible to add a reproducer ? Or too hard to replicate ?

@vrajat we need to extract some methods to write unit tests. This issue is causing some failure when parallel warmup is enabled.

Issue #220, opened 2018-10-17
Nonlocal request count is being reported as 0. Remote read request is increasing for subsequent runs


abhishekdas99:

stagraqubole:
stats have been refactored, this should be solved now

PR #221, opened 2018-10-17
fix: usr: #220: Create map to maintain cluster managers of different cluster types

abhishekdas99:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -645,6 +659,43 @@ public void onRemoval(RemovalNotification<String, FileMetadata> notification)
     }
   }
 
+  private class ClusterManagerMetadata
```
stagraqubole: Why this separate class? Isnt ClusterManager enough?

abhishekdas99: We need list of nodes and currentnodeindex. although these could be derived from clustermanager itself, (like `clustermanager.getNodes()` and `clustermanager.getNodes().getIndexOf(currentNodeName)`) thought of storing these info for use in other sections.

stagraqubole: Seems to be just bloating code as of now. Unless there is a need of it we can get rid of it.

PR #225, opened 2018-10-23
fix: dev: Remove redundant readData call to avoid extra download of files

abhishekdas99:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -213,10 +213,6 @@ public void run()
           throw new Exception("Could not create BookKeeper Client " + Throwables.getStackTraceAsString(e));
         }
 
-        if (!bookKeeperClient.readData(remotePath, offset, readLength, header.getFileSize(), header.getLastModified(), header.getClusterType())) {
-          throw new Exception("Could not cache data required by non-local node");
-        }
-
         if (!CacheConfig.isParallelWarmupEnabled(conf)) {
```
stagraqubole: Shouldnt you rather remove this if clause? With this change parallel warmup will not add anything to queue

abhishekdas99: Yes.. If parallel warmup is set to true, you call NonLocaReadrequest for the data which is already downloaded. So only invoke readdata (which in turn will put the request in the queue) if parallel warmup is not enabled.
I know this is a little convoltuted way of doing it and we will revisit this piece of code to make it easier to comprehend.

By the way, this change occurred due to some merging problem. Both the lines got merged instead of one

stagraqubole: Ok, I am forgetting the parallel warmup path. We need to have this info documented somewhere, parallel warmup related classes seems the best place to document this flow.

stagraqubole:
lgtm.

On related path, should we add a sanity test in ClientServiceThread.run() where it reads from file directly in case of reaching there in parallel warmup case? Due to some bug if we reach there without warmup, we could end up returning corrupt data in case of sparse file.

abhishekdas99:
What kind of sanity test we are looking for here . We can check the cache status before deciding to put the request the in the queue or read the file directly. The only drawback would be we will be making one more thrift call in that case

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
abhishekdas99: Agree to that. We want to do some refactoring around this area. We can definitely add documentation

stagraqubole:
> What kind of sanity test we are looking for here . We can check the cache status before deciding to put the request the in the queue or read the file directly. The only drawback would be we will be making one more thrift call in that case

If one extra thrift call, which is over loopback, isnt expensive then better do that than take a chance on corruption.

rubix-core/src/main/java/com/qubole/rubix/core/RemoteReadRequestChain.java
```diff
@@ -192,17 +192,28 @@ public ReadRequestChainStats getStats()
   @Override
   public void updateCacheStatus(String remotePath, long fileSize, long lastModified, int blockSize, Configuration conf)
   {
+    RetryingBookkeeperClient client = null;
     try {
-      RetryingBookkeeperClient client = bookKeeperFactory.createBookKeeperClient(conf);
+      client = bookKeeperFactory.createBookKeeperClient(conf);
```
jordanw-bq: You can use try-with-resources to make sure the client is automatically closed once the `try` block has finished.
ex: `try (RetryingBookKeeper client = bookKeeperFactory.createBookKeeperClient(conf)) {`

rubix-core/src/main/java/com/qubole/rubix/core/RemoteReadRequestChain.java
```diff
@@ -192,17 +192,28 @@ public ReadRequestChainStats getStats()
...
-      client.close();
     }
     catch (Exception e) {
       log.info("Could not update BookKeeper about newly cached blocks: " + Throwables.getStackTraceAsString(e));
     }
+    finally {
+      try {
+        if (client != null) {
+          client.close();
+          client = null;
```
jordanw-bq: `client` is only used in the scope of this function, so we shouldn't need to set it to null since it won't be used after this `finally` block

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -213,16 +217,30 @@ public void run()
...
+          long startBlock = offset / blockSize;
+          long endBlock = ((offset + (readLength - 1)) / blockSize) + 1;
+
+          CacheStatusRequest request = new CacheStatusRequest(remotePath, header.getFileSize(), header.getLastModified(),
+              startBlock, endBlock, header.getClusterType());
+          List<BlockLocation> blockLocations = bookKeeperClient.getCacheStatus(request);
+
+          for (BlockLocation location : blockLocations) {
+            if (location.getLocation() != Location.CACHED) {
+              throw new Exception("The requested data in not in cache. The data will be read from object store");
```
stagraqubole: will this be logged somewhere?

stagraqubole: lets log if not there anywhere

abhishekdas99: Done

Issue #186, opened 2018-07-25
Cobertura/Codecov not reporting correct numbers


jordanw-bq:
https://codecov.io/gh/qubole/rubix/pull/234/src/rubix-spi/src/main/java/com/qubole/rubix/spi/ClusterType.java

Coveralls does not seem to be reporting the correct coverage for our test cases.
Ex: In the report above, ClusterType dropped by almost 50% down to 0% coverage, even though this code was not touched as part of the PR.

Will require investigation into Travis build logs when generating Cobertura report

jordanw-bq:
Findings so far:

Test failures are due to Thrift classes not being shaded when tests are being run.
Cobertura runs the `test` phase to generate its report. However, shading happens in the `package` phase of the Maven build, which takes place after the `test` phase.

PR #246, opened 2018-12-03
fix: dev: #186: use appropriate Cobertura goal to shade jars before testing

jordanw-bq:
`mvn cobertura:cobertura-integration-test` runs the `verify` phase, which includes the `package` phase needed to shade the JARs for tests in later modules. This means that the `bookkeeper` tests should now all pass and report the correct coverage.


abhishekdas99:
LGTM

PR #252, opened 2018-12-18
fix: usr: #249: Fallback reading from object store if cached file is missing or corrupt

abhishekdas99:
This PR includes:
1. Fallback reading from object store if cached file is missing or corrupt
2. Moving core.utils classes (except DummyClusterManager) to common.utils
3. changed the logic in CacheUtil.getDirectory to work with path object rather than string manipulation. The existing implementation is not going to work for LocalFileSystem (required for unit tests)

The main classes to look at :
1. Implementation in rubix-core/src/main/java/com/qubole/rubix/core/CachedReadRequestChain.java
2. Tests in rubix-core/src/test/java/com/qubole/rubix/core/TestCachedReadRequestChain.java
3. Implementation in rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
4. Tests in rubix-spi/src/test/java/com/qubole/rubix/spi/TestCacheUtil.java


rubix-core/src/main/java/com/qubole/rubix/core/CachedReadRequestChain.java
```diff
@@ -74,44 +86,140 @@ public Integer call()
...
+        log.debug(String.format("CachedFileRead copied data [%d - %d] at buffer offset %d",
+                readRequest.getActualReadStart(),
+                readRequest.getActualReadStart() + nread,
+                readRequest.getDestBufferOffset()));
+
+        if (nread != readRequest.getActualReadLength()) {
+          needsInvalidation = true;
+          log.error(String.format("Cached read length didn't match with requested read length for file %s. " +
+                  " Falling back reading from object store.", localCachedFile));
+          directDataRead = readFromRemoteFileSystem(readRequests.indexOf(readRequest));
```
stagraqubole: why trust any data in that file? I suggest we just invalidate and return data from direct reads for all RRs in the chain

rubix-core/src/main/java/com/qubole/rubix/core/CachedReadRequestChain.java
```diff
@@ -74,44 +86,140 @@ public Integer call()
...
+      }
+      catch (IOException ex) {
+        log.error("Could not close bookkeeper client. Exception: " + ex.toString());
+      }
+    }
+  }
+
+  private int readFromRemoteFileSystem(int index) throws IOException
+  {
+    String remotePath = CacheUtil.getRemotePath(localCachedFile, conf);
```
stagraqubole: Code duplication with NonLocalReadRequestChain

rubix-core/src/main/java/com/qubole/rubix/core/CachedReadRequestChain.java
abhishekdas99: Done. Reading the entire request from remote object store.

PR #266, opened 2019-03-05
chg: fix: ISS-269: Fixed the security bugs reported

kvankayala:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -112,7 +112,7 @@ public void startServer(Configuration conf, MetricRegistry metricsRegistry)
       server.serve();
     }
     catch (TTransportException e) {
-      e.printStackTrace();
+      log.error(String.format(Throwables.getStackTraceAsString(e)));
```
jordanw-bq: Remove this line, since we are already logging the exception on the next line

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -542,7 +542,7 @@ private boolean readDataInternal(String remotePath, long offset, int length, lon
           fs.close();
         }
         catch (IOException e) {
-          e.printStackTrace();
+          log.error(String.format(Throwables.getStackTraceAsString(e)));
```
jordanw-bq: Remove `String.format()` on this and similar lines; this should only be used for formatting strings (ie. strings containing `%s` and the like)

PR #270, opened 2019-03-11
fix: usr: #260: Add metrics for cache available and download time

abhishekdas99:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloader.java
```diff
@@ -51,6 +51,7 @@
   int diskReadBufferSize;
   private MetricRegistry metrics;
   private Counter totalMBDownloaded;
+  private Counter totalTimeToDownload;
```
kvankayala: I think this **totalTimeToDownload** gives the cumulative value of total time spent in downloading all the data. How will this metric be useful to show any improvement in the query execution or cache utilization ?

abhishekdas99: This is useful to calculate the total time saved if data is read from the cache.

PR #275, opened 2019-03-14
chg: dev: #274: update Checkstyle version to address vulnerability

jordanw-bq:


jordanw-bq:
This change has been tested in my local environment, by changing the order of a class' imports and verifying that `mvn clean install` will properly alert that the order of the imports is incorrect.

PR #289, opened 2019-05-23
new: dev: ISS-288:Encrypt the local file path while downloading to local disk

kvankayala:
Currently, the downloaded files would be saved in local disk with same file path as it is in s3/wasb.
   To conceal the file paths stored in local, this PR introduces a configuration/feature **rubix.cache.path.encryption** which enables path encryption if set to true.


rubix-spi/src/test/java/com/qubole/rubix/spi/TestCacheUtil.java
```diff
@@ -69,6 +74,24 @@ public void setUpConfiguration()
     conf.clear();
   }
 
+  private static String getHashedPath(String relLocation)
```
jordanw-bq: This method should not be duplicated in the test class. You can mark the original method as `protected` and use the `@VisibleForTesting` annotation to signal the reason for the relaxed visibility.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -525,6 +532,11 @@ public static void setValidationEnabled(Configuration conf, boolean isValidation
     conf.setBoolean(KEY_VALIDATION_ENABLED, isValidationEnabled);
   }
 
+  public static void setpathEncryptionEnabled(Configuration conf, boolean isPathEncryptionEnabled)
```
jordanw-bq: Nit: fix camel-case (`setPath...`)

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -86,6 +86,7 @@
   private static final String KEY_MAX_CACHE_SIZE = "rubix.cache.max.size";
   private static final String KEY_CACHE_FILE_SPLIT_SIZE = "rubix.cache.filesplit.size";
   private static final String KEY_CLUSTER_NODE_REFRESH_TIME = "rubix.cluster.node.refresh.time";
+  private static final String KEY_ENABLE_PATH_ENCRYPTION = "rubix.cache.path.encryption";
```
jordanw-bq: Add `.enabled` to make it clearer that this is a key for a Boolean config option

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
```diff
@@ -244,13 +256,33 @@ private static String getDirectory(String remotePath, Configuration conf)
...
+      MessageDigest md = MessageDigest.getInstance("SHA-256");
+      byte[] pathBytes = md.digest(relLocation.getBytes());
+      StringBuilder sb = new StringBuilder();
+      for (int i = 0; i < pathBytes.length; i++) {
+        sb.append(Integer.toString((pathBytes[i] & 0xff) + 0x100, 16).substring(1));
+      }
+      hashRelLocation = sb.toString();
+    }
+    catch (NoSuchAlgorithmException e) {
+      log.error("No Such Algorithm for Hashing " + e.toString());
```
jordanw-bq: `log.error()` can take an exception as a second parameter. This is preferred, rather than concatenating the exception as a String.

rubix-spi/src/test/java/com/qubole/rubix/spi/TestCacheUtil.java
kvankayala: Done

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
kvankayala: Done

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
kvankayala: Done

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
kvankayala: Done

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeper.java
```diff
@@ -76,7 +76,7 @@ public void setUp() throws IOException
   {
     CacheConfig.setCacheDataDirPrefix(conf, TEST_CACHE_DIR_PREFIX);
     CacheConfig.setBlockSize(conf, TEST_BLOCK_SIZE);
-
+    CacheConfig.setPathEncryptionEnabled(conf, true);
```
jordanw-bq: Why are we enabling path encryption for these tests?

rubix-spi/src/test/java/com/qubole/rubix/spi/TestCacheUtil.java
```diff
@@ -39,6 +42,7 @@
 
 public class TestCacheUtil
 {
+  private static final Log log = LogFactory.getLog(CacheUtil.class);
```
jordanw-bq: Should be `TestCacheUtil.class`

kvankayala:
Introduced the configuration **rubix.cache.path.encryption.depth** and **rubix.cache.path.encryption.enabled** to define depth of the encryption(number of inner folders) and enable the file path encryption.

PR #290, opened 2019-05-31
Remove redundant pluginRepository from rubix-bookkeeper

garvit-gupta:


abhishekdas99:
This is already incorporated in thrift version 0.9.2
https://issues.apache.org/jira/browse/THRIFT-1536

PR #294, opened 2019-06-12
new: usr: #296: RubiX Dummy Mode

### LINKED ISSUE #296: Implement Dummy Mode
Implement Dummy Mode in which RubiX does not store any data in the cache locally, but just reads data directly from the source. Cache metadata is updated to maintain cache stats.

garvit-gupta:
- RubiX Dummy Mode would enable users to witness the impact of RubiX on query statistics such as Cached Reads, Hit Rate etc. without actually storing any data in the cache. 

- In Dummy Mode, RubiX reads the entire data directly from the source, while at the same time, Read Request Chains are setup. These chains are not called and thus never read any data, but they update the cache metadata for each read request.

fixes #296


garvit-gupta:
Testing Dummy Mode functionality:

- Verified that no data is getting stored in the cache (apart from mdfiles).
- Verified that Dummy mode Cache Stats match RubiX numbers, by comparing jmx numbers: Executed a tpcds query with and without dummy mode enabled. First it was executed thrice on orc_1000 and then thrice on parquet_1000. Recorded jmx query results after each of the 6 trials in both cases: https://docs.google.com/spreadsheets/d/1Cis_4R_jv6C8p02_o6ZIUMb1o9ni-wk9-5RDPNzvyAQ/edit?usp=sharing.
- Checked that there are no additional exceptions occurring in bks, lds and server logs with dummy mode.
- Verified correctness of query results by executing various sum, count and needle_in_haystack queries.
- Did a regression run (with the dummy mode code but with the feature disabled). There were no issues.
- Carried out a perf test by running 6 tpcds queries with dummy mode and with Presto without RubiX. Timings match.
DummyModePerf

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -382,7 +393,7 @@ else if (isCached.get(idx).getLocation() == Location.CACHED) {
         catch (IOException e) {
           log.error("Unable to open file channel in R mode", e);
           // reset bookkeeper client so that we take direct route
-          this.bookKeeperClient = null;
+          bookKeeperClient = null;
```
stagraqubole: This will not do the intended work, setting argument to null will not set the value null at caller.

rubix-core/src/main/java/com/qubole/rubix/core/DummyModeCachingInputStream.java
```diff
@@ -0,0 +1,95 @@
...
+    return getParentDataInputStream().getPos();
+  }
+
+  @Override
+  public void seek(long pos)
+          throws IOException
+  {
+    checkState(pos >= 0, "Negative Position");
+    long currentPos = getPos();
+    log.debug(String.format("Dummy Mode: Seek request, currentPos: %d currentBlock: %d", currentPos, currentPos / blockSize));
```
stagraqubole: remove log

rubix-core/src/main/java/com/qubole/rubix/core/DummyModeCachingInputStream.java
```diff
@@ -0,0 +1,95 @@
...
+    getParentDataInputStream().seek(pos);
```
stagraqubole: Keep function content only line 62

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -81,7 +82,7 @@
   private byte[] affixBuffer;
   private int diskReadBufferSize;
   private int bufferSize;
-  private BookKeeperFactory bookKeeperFactory;
+  BookKeeperFactory bookKeeperFactory;
```
stagraqubole: This can be private

rubix-core/src/main/java/com/qubole/rubix/core/DummyModeCachingInputStream.java
```diff
@@ -0,0 +1,95 @@
...
+
+    readService.execute(new Runnable()
+    {
+      @Override
+      public void run()
+      {
+        try {
+          long endBlock = ((initPos + (length - 1)) / blockSize) + 1;
+          final List<ReadRequestChain> readRequestChains = setupReadRequestChains(buffer, offset, endBlock, length, initPos, initNextReadBlock, bookKeeperFactory.createBookKeeperClient(conf));
+          updateCacheAndStats(readRequestChains);
```
stagraqubole: `updateCacheAndStats` will submit a work on readService. We dont need this redirection, the current work is happening on readService thread so let the updation also happen there. You can change `updateCacheAndStats` to do only the actual work, line299 to line305, and let the callers call it in a executor thread

rubix-core/src/main/java/com/qubole/rubix/core/NonLocalReadRequestChain.java
```diff
@@ -193,4 +197,45 @@ private int directReadRequest(int index)
...
+  public void updateCacheStatus(String remotePath, long fileSize, long lastModified, int blockSize, Configuration conf)
+  {
+    if (CacheConfig.isDummyModeEnabled(conf)) {
+      RetryingBookkeeperClient bookKeeperClient = null;
+      try {
+        bookKeeperClient = new BookKeeperFactory().createBookKeeperClient(remoteNodeName, conf);
+        for (ReadRequest readRequest : readRequests) {
+          long startBlock = toBlock(readRequest.getBackendReadStart());
+          long endBlock = toBlock(readRequest.getBackendReadEnd() - 1) + 1;
+          log.debug("Dummy Mode: Non-local Read Request: Setting cached from : " + startBlock + " block to : " + endBlock);
```
stagraqubole: remove logs.

rubix-core/src/main/java/com/qubole/rubix/core/NonLocalReadRequestChain.java
```diff
@@ -193,4 +197,45 @@ private int directReadRequest(int index)
...
+          CacheStatusRequest request = new CacheStatusRequest(remotePath, fileSize, lastModified, startBlock, endBlock, clusterType);
+          bookKeeperClient.getCacheStatus(request);
```
stagraqubole: Add a comment why getCacheStatus is needed

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
stagraqubole: Doing this was not the right thing to do at the first place, we are not closing the underlying transport which can cause leaks. 
We set `bookKeeperClient = null` to force next rounds of reads to fallback to direct reads when we see FS permission issues in opening files. I suggest we change `bookKeeperClient = null;` logic to instead use a variable in the CachingInputStream to mark permission errors, against which we would fallback to direct reads.

rubix-core/src/main/java/com/qubole/rubix/core/DummyModeCachingInputStream.java
garvit-gupta: Okay will do

rubix-core/src/main/java/com/qubole/rubix/core/NonLocalReadRequestChain.java
garvit-gupta: Okay

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
garvit-gupta: Will raise an issue for this

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
garvit-gupta: Required in DummyModeCachingInputStream

rubix-core/src/main/java/com/qubole/rubix/core/NonLocalRequestChain.java
```diff
@@ -165,4 +166,45 @@ public Integer call() throws Exception
 
     return nonLocalReadBytes;
   }
+
+  @Override
+  public void updateCacheStatus(String remotePath, long fileSize, long lastModified, int blockSize, Configuration conf)
+  {
+    if (CacheConfig.isDummyModeEnabled(conf)) {
+      if (remoteFetchRequestChain == null || remoteFetchRequestChain.getReadRequests().isEmpty()) {
```
abhishekdas99: We should just call remoteFetchRequestChain.updateCacheStatus for each of the remoteFetch request chains. This will get rid of the duplication of the code here.

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -245,7 +246,10 @@ private int readInternal(byte[] buffer, int offset, int length)
     final List<ReadRequestChain> readRequestChains = setupReadRequestChains(buffer,
         offset,
         endBlock,
-        length);
+        length,
+        nextReadPosition,
```
abhishekdas99: Instead of taking `nextReadPosition` and `nextReadBlock`, can we just make this two class variables non private. We are passing the same value that is set in CachingInputStream

abhishekdas99:
@garvit-gupta  Can you also change the configuration.rst file to include the newly added configuration for dummy mode.

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
stagraqubole: We cant do this because setupReadRequestChains on same InputStream in dummy mode can be called in parallel by multiple threads because we offload it to executor in every read.  Reusing the class members causes issues because threads now step over each other, hence dummy InputStream stores the required variables locally and pass it to this function.

rubix-core/src/main/java/com/qubole/rubix/core/DummyModeCachingInputStream.java
```diff
@@ -0,0 +1,92 @@
...
+    final int read = readFullyDirect(buffer, offset, length);
+
+    readService.execute(new Runnable()
+    {
+      @Override
+      public void run()
+      {
+        try {
+          long endBlock = ((initPos + (length - 1)) / blockSize) + 1;
+          final List<ReadRequestChain> readRequestChains = setupReadRequestChains(buffer, offset, endBlock, length, initPos, initNextReadBlock, bookKeeperFactory.createBookKeeperClient(conf));
```
abhishekdas99: we are creating a new bookkeeperclient object every time but we are closing that object. This will cause connection leak

abhishekdas99:
@garvit-gupta I have raised a pull request https://github.com/qubole/rubix/pull/309 where we have removed some redundant bookkeeper null set. Can you  merge that in your PR once its merged. With this change, some of the changes you did are not required.

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
abhishekdas99: @garvit-gupta  we are passing the `bookKeeperFactory` object in DummyModeCachingInputStream, so do we still need to be protected ?

PR #295, opened 2019-06-12
fix: dev: #297: Fix No such file or directory exception in bks logs

### LINKED ISSUE #297: Cache data directories to be made writable for the creation of metadata files

garvit-gupta:
The exception occurred because the bks service did not have write privileges over the cache table directories and consequently was not able to create metadata files in the case of Non local reads. 

- Making the cached table directories writable recursively from Cache util

fixes #297

### DIFF
--- a/rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
+++ b/rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
@@ -57,7 +57,7 @@ public class CacheUtil
         if (exists(cacheParentDir)) {
           parentDirectoryExists = true;
           final String cacheDirPath = cacheParentDir + "/" + CacheConfig.getCacheDataDirSuffix(conf);
-          createCacheDirectory(cacheDirPath);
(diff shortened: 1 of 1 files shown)


rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
```diff
@@ -208,7 +208,11 @@ private static void createCacheDirectory(String cacheDirPath)
   {
     final File cacheDir = new File(cacheDirPath);
     cacheDir.mkdirs();
-    cacheDir.setWritable(true, false);
+    File parentFile = cacheDir;
+    do {
+      parentFile.setWritable(true, false);
+      parentFile = parentFile.getParentFile();
+    } while (parentFile.getAbsolutePath().contains("fcache"));
```
abhishekdas99: Remove "fcache" and get it from the config

garvit-gupta: Made the change

PR #306, opened 2019-06-24
fix: dev: ISS-298: Fixed and improved debug logs 

kvankayala:
1. Printing file name in "Got Read" log.
2. Printing proper buffer name suffix/prefix/destination in "Trying to copy" log line.
3. Printing proper blockSize in BookKeeper.java:549 instead of **startBlock**
4. We are setting child thread-name as parent thread-name for readRequestChain and it's respective child threads. Instead, we can print parent-child thread-name for better debugging.


stagraqubole:
Through this, can you also take up the work to change logs which swallow exceptions and use the correct logger methods to log stack traces.

abhishekdas99:
> Through this, can you also take up the work to change logs which swallow exceptions and use the correct logger methods to log stack traces.

+1 for that.

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -99,7 +99,7 @@ public CachingInputStream(FileSystem parentFs, Path backendPath, Configuration c
       this.lastModified = fileInfo.lastModified;
     }
     catch (Exception ex) {
-      log.error("Could not get FileInfo for " + backendPath.toString() + " . Fetching FileStatus from remote file system");
+      log.error("Could not get FileInfo for " + backendPath.toString() + " . Fetching FileStatus from remote file system \n StackTrace : ", ex);
```
jordanw-bq: Having the exception as the second argument of the `log.error` call should automatically put the stack trace on a new line.
Also, I feel like we shouldn't need to specify that it is a stack trace; it should be clear enough from what is logged.

Issue #307, opened 2019-06-27
Add configuration option for server socket timeout


jordanw-bq:
Currently, the same timeout is used for connecting to & communicating with the BookKeeper server. This causes issues when trying to reduce connection timeouts for retrying client connections, since there are some BookKeeper operations that can take longer than this timeout.

Specifying separate timeouts for connection and socket communication will allow more flexibility for configuring communication with the BookKeeper server.

PR #308, opened 2019-06-27
chg: usr: #307: add configuration option for server socket timeout

jordanw-bq:
* Tested & verified working on Qubole cluster, where we were experiencing this issue


rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -90,6 +91,7 @@
   // default values
   private static final int DEFAULT_BLOCK_SIZE = 1 * 1024 * 1024; // 1MB
   private static final int DEFAULT_SERVER_CONNECT_TIMEOUT = 1000; // ms
+  private static final int DEFAULT_SERVER_SOCKET_TIMEOUT = 3000; // ms
```
abhishekdas99: I would suggest to make the default value to 6 sec as some of the initial S3 calls are taking 2-3 sec. So to be in the safe side, lets make it 6sec

abhishekdas99:
@jordanw-bq Is it possible to add couple of  simple unit tests covering the cases where you are getting a timeout and not getting a timeout exception.

rubix-spi/src/test/java/com/qubole/rubix/spi/TestBookKeeperFactory.java
```diff
@@ -0,0 +1,195 @@
...
+    final int aliveCallDelay = 1000;
+    final int socketTimeout = 500;
+    final int connectTimeout = 500;
+
+    startMockServer(true, NO_DELAY, aliveCallDelay);
+
+    final RetryingBookkeeperClient client = createTestBookKeeperClient(socketTimeout, connectTimeout);
+    client.isBookKeeperAlive(); // should throw expected exception due to socket timeout
+  }
+
```
abhishekdas99: Can we add a test where we are able to connect although there is a delay in the server ( delay is less than the socket time out). In that case we are not going to get any exception

jordanw-bq: @abhishekdas99 Added tests to cover more cases

Issue #281, opened 2019-04-04
Remove clusterType reference from client side and Worker Bookkeeper


abhishekdas99:
This is dependent on #188 . As part of #188 we are initializing the cluster manager object from Coordinator Bookkeeper so we need to remove reference of clusterType from elsewhere.

stagraqubole:
Original change reverted, closing this.

PR #330, opened 2019-10-16
fix: dev: #281: Remove clusterType reference from client side and Worker Bookkeeper

abhishekdas99:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -50,8 +47,7 @@
   private BookKeeperFactory bookKeeperFactory;
   // The hostname of the master node.
   private String masterHostname;
-  private int clusterType;
-  private RetryingBookkeeperClient client;
+  RetryingBookkeeperClient client;
```
kvankayala: nit: please explicitly mention `private` access modifier to maintain uniformity

PR #335, opened 2020-01-09
fix: dev: ISS-334: ReadRequestChain threads got deamonized

kvankayala:
Daemonized the ReadRequestChain threads


abhishekdas99:
@kvankayala  Can we create another issue to make sure we tackle this non-daemonized thread issue across the project.

PR #338, opened 2020-02-03
Changes for embedded mode

stagraqubole:
Following changes have been done:
- Start LTS with BookKeeper object
- Use Bookkeeper object in ReadRequestChains
- Use Bookkeeper object in parallel warmup
- Fix Parallel warmup for the case of CachingFS getting used by disabling caching on configuration used in parallel warmup
- Fix logging: Fix log levels, show full exceptions whereever available instead of exception message
- Recover from loss of connection to master BK in workerBK
- Use prefix+suffix cache location if count based locations not found
- Add new module 'rubix-prestosql' for prestosql versions


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
```diff
@@ -119,7 +119,7 @@ protected ClusterManager getClusterManagerInstance(ClusterType clusterType, Conf
           throws CoordinatorInitializationException
   {
     String clusterManagerClassName = CacheConfig.getClusterManagerClass(conf, clusterType);
-    log.info("Initializing cluster manager : " + clusterManagerClassName + " for cluster type " + clusterType);
+    log.debug("Initializing cluster manager : " + clusterManagerClassName + " for cluster type " + clusterType);
```
kvankayala: I think it is okay to good to keep this log as **info** as we see it only once when the Bks gets started. And, it sometimes helps in debugging as well.

pom.xml
```diff
@@ -62,11 +63,13 @@
         <module>rubix-client</module>
         <module>rubix-common</module>
         <module>rubix-presto</module>
+        <module>rubix-prestosql</module>
         <module>rubix-hadoop2</module>
         <module>rubix-bookkeeper</module>
         <module>rubix-tests</module>
         <module>assembly</module>
         <module>rubix-rpm</module>
+        <module>rubix-shaded</module>
```
kvankayala: module name could be more specific i.e., **rubix-prestosql-shaded**

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -82,9 +83,32 @@ public void run()
...
+    else {
+      startThriftServer(conf, localBookKeeper);
+    }
+
+    return localBookKeeper;
+  }
+
+  public void setupServer(Configuration conf, MetricRegistry metricsRegistry)
+  {
+    conf = new Configuration(conf);
```
kvankayala: Why are we creating new conf object?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -82,9 +83,32 @@ public void run()
...
+    CacheConfig.setCacheDataEnabled(conf, false);
```
kvankayala: Shouldn't we disable Rubix only if it is in embeddedMode?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
```diff
@@ -99,14 +123,15 @@ public void startServer(Configuration conf, MetricRegistry metricsRegistry)
...
+      throw Throwables.propagate(e);
     }
-
-    startThriftServer(conf, localBookKeeper);
   }
 
   void startServer(Configuration conf, BookKeeper bookKeeper, BookKeeperMetrics bookKeeperMetrics)
   {
+    conf = new Configuration(conf);
+    CacheConfig.setCacheDataEnabled(conf, false);
```
kvankayala: same here

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/CoordinatorBookKeeper.java
stagraqubole: We can enable debug level logs while triaging ... this sort of problems should mostly happen during our testing and changing log levels are easy at that time.

pom.xml
stagraqubole: I avoided that because prestodb version will be deprecated slowly. But prestosql here would avoid confusion about the contents, will refactor.

stagraqubole: I thought I had changed it, my comments was thinking it is named rubix-presto-shaded. Will rename to that.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
stagraqubole: This is disabling usage of cache inside the bookkeeperserver if somehow we end up with CachingFS object for FileSystem.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
stagraqubole: Same explanation. We dont need to use CachingFileSystem ever inside BookKeeper or LDS

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeperServer.java
stagraqubole: So that any changes made in the conf object inside BKS are not reflected in client side. In case of embedded mode with FileSystem cache enabled, this can disable Rubix for client if we use the same conf object passed down.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
```diff
@@ -0,0 +1,289 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Created by stagra on 14/1/16.
+ */
+public class PrestoClusterManager extends ClusterManager
+{
+  private int serverPort = 8081;
```
kvankayala: Make 8081 as configurable.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
```diff
@@ -0,0 +1,289 @@
...
+  private String serverAddress = "localhost";
+  static LoadingCache<String, List<ClusterNode>> nodesCache;
+
+  private Log log = LogFactory.getLog(PrestoClusterManager.class);
+
+  public static String serverPortConf = "caching.fs.presto-server-port";
+  public static String serverAddressConf = "master.hostname";
```
kvankayala: "master.hostname" can be obtained from CacheConfig.getCoordinatorHostName(conf);

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
```diff
@@ -0,0 +1,289 @@
...
+  public static String yarnServerAddressConf = "yarn.resourcemanager.address";
```
kvankayala: Similarly, getResourceManagerAddress(conf) will provide yarn server address.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
```diff
@@ -0,0 +1,289 @@
...
+
+              int allNodesResponseCode = allHttpCon.getResponseCode();
+
+              StringBuffer allResponse = new StringBuffer();
+              StringBuffer failedResponse = new StringBuffer();
+              try {
+                if (allNodesResponseCode == HttpURLConnection.HTTP_OK) {
+                  BufferedReader in = new BufferedReader(new InputStreamReader(allHttpCon.getInputStream()));
+                  String inputLine = "";
+                  try {
```
kvankayala: You can use **try-with-resources** instead of try-catch-finally for closeable inputstream **in**

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
```diff
@@ -0,0 +1,289 @@
...
+              finally {
+                allHttpCon.disconnect();
+              }
+
+              HttpURLConnection failHttpConn = (HttpURLConnection) failedNodesRequest.openConnection();
+              failHttpConn.setConnectTimeout(500);    //ms
+              failHttpConn.setRequestMethod("GET");
+              int failedNodesResponseCode = failHttpConn.getResponseCode();
+              // check on failed nodes
+              try {
```
kvankayala: Same here! You can use **try-with-resources** instead of try-catch-finally

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/CachingPrestoS3FileSystem.java
```diff
@@ -0,0 +1,44 @@
...
+ * Created by stagra on 14/1/16.
+ */
+public class CachingPrestoS3FileSystem extends CachingFileSystem<PrestoS3FileSystem>
+{
+  public CachingPrestoS3FileSystem()
+  {
+    super();
+  }
+
+  private static final String SCHEME = "s3n";
```
kvankayala: I think the SCHEME should be **s3a**. Correct me if I'm wrong.

rubix-prestosql/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.7</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>19.0</version>
```
kvankayala: Main **pom.xml** includes 18.0 version of the guava. And, **rubix-presto** and **rubix-prestosql** are using 19.0. I think it is good to use a consistent version across the library. So, Can you please use **<version>${dep.guava.version}</version>** in dependency. Please let me know if 18.0 is not compatible with 19.0

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
stagraqubole: Already done, check setPrestoServerPort

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
stagraqubole: Yes, old code, not used. This is carried over from rubix-presto code.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
stagraqubole: Same explanation, will be removing both but only in rubix-prestosql

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
stagraqubole: all old code hence the old standards. We need to modernize the whole project.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/CachingPrestoS3FileSystem.java
stagraqubole: old code again, it is from the changes to support rubix scheme, i think it doesnt matter if we use s3n or s3a here as old systems maps them to a FS.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
kvankayala: Sorry, This comment applies to below **try-catch** block for inputstream **in**

rubix-prestosql/pom.xml
stagraqubole: Right, versioning is all over the place. Bookkeeper and core use guava18 while hadoop2,presto,prestosql are using 19. We should fix it but I would keep that a separate change, will open a jira for it.

stagraqubole: https://github.com/qubole/rubix/issues/340

rubix-prestosql/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+        <dependency>
+            <groupId>com.qubole.rubix</groupId>
+            <artifactId>rubix-common</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-common</artifactId>
+            <scope>provided</scope>
```
kvankayala: Does presto has **hadoop-common** jars available in presto plugins? How can rubix get hadoop jars if they run in presto JVM ?

stagraqubole: they are there in presto.

PR #339, opened 2020-02-06
Allow connection backlog for socket accept()

stagraqubole:
In a test program to make 1000 BKS calls via 100 threads, lot of calls failed with "Connection reset" and other Socket Exceptions. It happens because the way we are configuring thrift server, it does not queue up any connection while socket accept() is busy handing over work to a worker thread. This change enables queueing up such connections by setting the queue length to maximum by setting it to Integer.MAX_VALUE (the real limit is implementation specific and not documented anywhere, used Integer.MAX_VALUE to get the maximum queue length).
After this change I could not repro the socket exceptions in my test program even after increasing number of calls and number of threads.


kvankayala:
@stagraqubole In case of spark driver in worker nodes, I could see a lot of **connect timed out** and **Read timed out** like 250 and 100 respectively.  It's happening because of master bks getting overloaded with connection requests. I think increasing socket timeout and read timeout would also help. 
cc: @abhishekdas99

PR #344, opened 2020-03-23
Adding connection pool for BKS

harmandeeps:
Changes:
1. Added fast object pool.
2. Changes to accommodate pool in Rubix.
3. Added perf utility to run parallel queries on Presto server.

TODO:
1. Setting pool properties with config file.
2. Refactor the code


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -412,7 +412,7 @@ public FileInfo getFileInfo(String remotePath) throws TException
         return info;
       }
       catch (Exception e) {
-        log.error(String.format("Could not fetch FileStatus from remote file system for %s : %s", remotePath), e);
+        log.error(String.format("Could not fetch FileStatus from remote file system for %s : %s", remotePath, e));
```
stagraqubole: keep `e` an argument to log.error. Change the string format instead.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -176,10 +176,10 @@ private void initializeNodeStateCache(final Configuration conf, final Ticker tic
               return client.getClusterNodes();
             }
             catch (Exception e) {
-              client.close();
               throw e;
             }
             finally {
+              bookKeeperFactory.returnBookKeeperClient(client.getTransportPoolable());
```
stagraqubole: Instead of each user of BookKeeperClient to be aware of pooling keep the pooling work limited to BookKeeperClient. Continue to have BookKeeperClient.close() which internally releases the transport.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
harmandeeps: fixed.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
harmandeeps: moved logic in close method and user try with auto closeable wherever needed.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/WorkerBookKeeper.java
```diff
@@ -176,10 +176,10 @@ private void initializeNodeStateCache(final Configuration conf, final Ticker tic
               return client.getClusterNodes();
             }
             catch (Exception e) {
```
stagraqubole, line 178: delete this catch

rubix-spi/pom.xml
```diff
@@ -44,6 +44,13 @@
             <version>${dep.mockito.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.conversantmedia</groupId>
+            <artifactId>disruptor</artifactId>
+            <version>1.2.14</version>
+            <!-- This is optional, if you want the best performance you need to manually add disruptor into your dependency -->
```
stagraqubole, line 51: drop the comment

rubix-spi/src/main/java/com/qubole/rubix/spi/BookKeeperFactory.java
```diff
@@ -20,13 +24,84 @@
...
+    poolConfig.setMaxIdleMilliseconds(60 * 1000 * 5);
+    final int socketTimeout = 60000;
+    final int connectTimeout = 10000;
+
+    factory = new ObjectFactory<TSocket>()
+    {
+      @Override
+      public TSocket create(String host, int socketTimeout, int connectTimeout)
+      {
+        log.debug("Opening connection on host: " + host);
```
stagraqubole, line 64: "Opening connection on host" -> "Opening connection to host"

rubix-spi/src/main/java/com/qubole/rubix/spi/BookKeeperFactory.java
```diff
@@ -20,13 +24,84 @@
...
+        TSocket socket = null;
+        try {
+          socket = new TSocket(host, 8899, socketTimeout, connectTimeout); // create your object here
```
stagraqubole, line 67: drop comment

rubix-spi/src/main/java/com/qubole/rubix/spi/BookKeeperFactory.java
```diff
@@ -20,13 +24,84 @@
...
+
 /**
  * Created by sakshia on 5/10/16.
  */
 public class BookKeeperFactory
 {
   BookKeeperService.Iface bookKeeper;
   private static Log log = LogFactory.getLog(BookKeeperFactory.class.getName());
+  private static final AtomicInteger count = new AtomicInteger();
+  private static final AtomicBoolean flag = new AtomicBoolean();
```
stagraqubole, line 40: rename flag to `initLock`

rubix-spi/src/main/java/com/qubole/rubix/spi/BookKeeperFactory.java
```diff
@@ -20,13 +24,84 @@
...
+  private static PoolConfig poolConfig;
+  private static ConcurrentHashMap<String, Integer> concurrentHashMap = new ConcurrentHashMap<>();
```
stagraqubole, line 42: Rename `concurrentHashMap` to hostToPartitionId

rubix-spi/src/main/java/com/qubole/rubix/spi/BookKeeperFactory.java
```diff
@@ -20,13 +24,84 @@
...
+  private static ObjectFactory<TSocket> factory;
+  static ObjectPool pool;
+
+  private static final String LOCALHOST = "localhost";
+
+  static {
+    // TODO: add config.properties file to set this config
+    poolConfig = new PoolConfig();
+    poolConfig.setPartitionSize(10);
```
stagraqubole, line 51: partition size limits the number of hosts we can have in a cluster. We shouldnt have this limit, instead grow the partitions in ObjectPool dynamically during registerHost call if index is out of bounds

rubix-spi/src/main/java/com/qubole/rubix/spi/BookKeeperFactory.java
```diff
@@ -20,13 +24,84 @@
...
+    poolConfig.setMaxSize(2000);
+    poolConfig.setMinSize(100);
+    poolConfig.setDelta(100);
+    poolConfig.setMaxIdleMilliseconds(60 * 1000 * 5);
+    final int socketTimeout = 60000;
```
stagraqubole, line 56: Timeouts should be picked from configuration objects. But to do that we need access to configuration object which might not available during static initialization.
To solve this, we can move all this initialization to lazy processing. When there is a first call to create bookkeeper client, during that we do this initialization.

rubix-spi/src/main/java/com/qubole/rubix/spi/BookKeeperFactory.java
```diff
@@ -39,23 +114,39 @@ public BookKeeperFactory(BookKeeperService.Iface bookKeeper)
     }
   }
 
-  public RetryingBookkeeperClient createBookKeeperClient(String host, Configuration conf) throws TTransportException
+  public RetryingPooledBookkeeperClient createBookKeeperClient(String host, Configuration conf) throws TTransportException
   {
     if (bookKeeper != null) {
       return new LocalBookKeeperClient(null, bookKeeper);
```
stagraqubole, line 120: This needs a cleanup, we can take it up separately. We should return localBookKeeper only for Localhost and for other make connection and have a single bookkeeperfactory. Currently factory is instantiated at multiple places

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPool.java
```diff
@@ -0,0 +1,171 @@
...
+  public Poolable<T> borrowObject(int partitionNumber)
+          throws SocketException
+  {
+    return borrowObject(true, partitionNumber);
+  }
+
+  public Poolable<T> borrowObject(boolean blocking, int partitionNumber)
+          throws SocketException
+  {
+    Log.info("Borrowing object for partition: ", partitionNumber);
```
stagraqubole, line 65: change log level to debug for all pooling classes unless error/warnings are logged

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPool.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+    throw new RuntimeException("Cannot find a valid object");
+  }
+
+  private Poolable<T> getObject(boolean blocking, int partitionNumber)
+  {
+    if (shuttingDown) {
+      throw new IllegalStateException("Your pool is shutting down");
+    }
+    int partition = partitionNumber % this.config.getPartitionSize();
```
stagraqubole, line 83: '%' in our case can be fatal. Instead just use partitionNumber as is and checkState that partitionNumber < numberOfPartitions

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPool.java
```diff
@@ -0,0 +1,171 @@
...
+            throw new RuntimeException("Cannot get a free object from the pool");
+          }
+        }
+      }
+      catch (InterruptedException e) {
+        throw new RuntimeException(e); // will never happen
+      }
+    }
+    else {
+      freeObject = subPool.getObjectQueue().poll();
```
stagraqubole, line 107: There is a race here due to which we can return 'null' and hit NPE up in the stack.
When `subPool.getObjectQueue().size()` was checked, size was non-zero so we came to else but on execution of `freeObject = subPool.getObjectQueue().poll();` we can get null if another thread in parallel did a poll.

We should refactor this method to this:
`
private Poolable<T> getObject(boolean blocking, int partition)
  {
    if (shuttingDown) {
      throw new IllegalStateException("Your pool is shutting down");
    }
    ObjectPoolPartition<T> subPool = this.partitions[partition];

    Poolable<T> freeObject;
    if (subPool.getObjectQueue().size() == 0) {
      // increase objects and return one, it will return null if reach max size
      subPool.increaseObjects(this.config.getDelta());
    }
    
    try {
      if (blocking) {
        Log.warn("Pool is empty...Waiting for connection");
        freeObject = subPool.getObjectQueue().take();
      }
      else {
        freeObject = subPool.getObjectQueue().poll(config.getMaxWaitMilliseconds(), TimeUnit.MILLISECONDS);
        if (freeObject == null) {
          throw new RuntimeException("Cannot get a free object from the pool");
        }
      }
    }
    catch (InterruptedException e) {
      throw new RuntimeException(e); // will never happen
    }
    
    freeObject.setLastAccessTs(System.currentTimeMillis());
    return freeObject;
  }
`

rubix-spi/src/main/java/com/qubole/rubix/spi/RetryingPooledBookkeeperClient.java
```diff
@@ -28,23 +29,32 @@
...
 {
-  private static final Log log = LogFactory.getLog(RetryingBookkeeperClient.class);
+  private static final Log log = LogFactory.getLog(RetryingPooledBookkeeperClient.class);
   private int maxRetries;
   TTransport transport;
+  Poolable<TTransport> transportPoolable;
 
-  public RetryingBookkeeperClient(TTransport transport, int maxRetries)
+  public RetryingPooledBookkeeperClient(TTransport transport, int maxRetries)
   {
```
stagraqubole, line 44: remove this constructor. The only user is LocalBookKeeperClient and it could be updated to use Poolable<TTransport> constructor

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/DisruptorObjectPool.java
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License. See accompanying LICENSE file.
+ * <p>
+ * <p>
+ * NOTICE: THIS FILE HAS BEEN MODIFIED BY  Qubole Inc UNDER COMPLIANCE WITH THE APACHE 2.0 LICENCE FROM THE ORIGINAL WORK
+ * OF https://github.com/DanielYWoo/fast-object-pool.
+ */
+package com.qubole.rubix.spi.fop;
```
stagraqubole, line 17: This calls is unused and can be removed. With this the dependency on 'com.conversantmedia.util' can also be removed

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/Log.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * @author Daniel
+ */
+public class Log
```
stagraqubole, line 27: Remove this, instead of this use LogFactory as used in other classes

rubix-core/src/main/java/com/qubole/rubix/core/CachedReadRequestChain.java
```diff
@@ -176,25 +176,12 @@ public void cancel()
 
   private void invalidateMetadata()
   {
-    RetryingBookkeeperClient client = null;
-    try {
-      client = factory.createBookKeeperClient(conf);
+    try (RetryingPooledBookkeeperClient client = factory.createBookKeeperClient(conf)) {
       client.invalidateFileMetadata(remotePath);
     }
     catch (Exception e) {
```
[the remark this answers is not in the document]
stagraqubole: This is handled, the client close() just does a put now so shouldnt be throwing any exceptions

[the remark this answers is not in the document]
harmandeeps: But, here `invalidateFileMetadata` throws an exception, that'ss why I didn't remove it.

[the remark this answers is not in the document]
stagraqubole: Yes, never mind this. I assumed that client.close() can also throw an exception which would be handled by this catch and the message for it would be wrong as invalidateFileMetadata did pass correctly. As mentioned later, it is handled so this not a issue.

rubix-spi/src/main/java/com/qubole/rubix/spi/BookKeeperFactory.java
harmandeeps: added logic of keeping Map<Host, Object Pool>, this way we can directly deal with the host and remove partition index logic.

rubix-spi/src/main/java/com/qubole/rubix/spi/RetryingPooledBookkeeperClient.java
harmandeeps: this constructor is used in many tests and creating object of transportpoolable is a little messy, so added annotation of 'VisibleOnlyForTesting'.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPool.java
```diff
@@ -0,0 +1,132 @@
...
+
+    Poolable<T> freeObject;
+    if (subPool.getObjectQueue().size() == 0) {
+      // increase objects and return one, it will return null if reach max size
+      subPool.increaseObjects(this.config.getDelta());
+    }
+
+    try {
+      if (blocking) {
+        log.warn("Pool is empty...Waiting for connection");
```
stagraqubole, line 94: Given this will be used in every call, warn level if too high. Infact, remote this log as take() doesnt necessarily mean waiting

harmandeeps: yeah, this is an old log, we should remove this.

PR #346, opened 2020-04-06
Fixes over connection pooling 

stagraqubole:
Main fixes:

```
ObjectPool
	Returns exception when no connections could be created
	
ObjectPoolPartition
	During creating min-connections, only add ones to the queue that are opened successfully
	During increaseObjects, only add ones to the queue that are opened successfully

RetryingPooledBookkeeperClient
	implements BookKeeperService.Iface and cover all underlying thrift functions
	throws away Pooled transport (after closing it) when it sees exception in making thrift calls and opens new one

BookKeeperFactory.factory to not return unopened sockets. Returns null now instead of bad socket

Test fixes:
BookKeeperFactory.resetConnectionPool() to reset pool to prevent connection reuse across testcases

Left:
	Get rid of `RetryingPooledBookkeeperClient()` constructor and across codebase use BookKeeperService.Iface reference instead of RetryingPooledBookkeeperClient
```


rubix-tests/src/test/java/com/qubole/rubix/tests/TestNonLocalReadRequestChain.java
```diff
@@ -323,13 +324,16 @@ private void stopAllServers()
     if (bookKeeperServer != null) {
       bookKeeperServer.stopServer();
     }
-    LocalDataTransferServer.stopServer();
+    if (LocalDataTransferServer.isServerUp()) {
```
harmandeeps, line 327: can't we add this condition LocalDataTransferServer.isServerUp() in isServerUp() method itself?

rubix-spi/src/main/java/com/qubole/rubix/spi/RetryingPoolingClient.java
```diff
@@ -0,0 +1,91 @@
...
+        return callable.call();
+      }
+      catch (Exception e) {
+        log.warn("Error while connecting : ", e);
+        errors++;
+        // We dont want to keep the transport around in case of exception to prevent reading old results in transport reuse
+        if (client.getInputProtocol().getTransport().isOpen()) {
+          // Close connection and submit back so that ObjectPool replaces it with new connection later
+          client.getInputProtocol().getTransport().close();
+          transportPoolable.getPool().returnObject(transportPoolable);
```
harmandeeps, line 73: why are we returning transportPoolable here? should we remove it from the pool?

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPool.java
```diff
@@ -86,8 +86,8 @@ public void registerHost(String host, int socketTimeout, int connectTimeout)
     Poolable<T> freeObject;
     if (subPool.getObjectQueue().size() == 0) {
       // increase objects and return one, it will return null if reach max size
-      int added = subPool.increaseObjects(this.config.getDelta());
-      if (added == 0) {
+      int totalObjects = subPool.increaseObjects(this.config.getDelta());
```
harmandeeps, line 89: so now we are not checking if totalObjects>0 and the additional connections are not created successfully?

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -855,6 +855,11 @@ public static void setTranportPoolMinSize(Configuration conf, int count)
     conf.setInt(KEY_POOL_MIN_SIZE, count);
   }
 
+  public static void setTranportPoolMaxSize(Configuration conf, int count)
```
harmandeeps, line 858: I created getters with name like 'getPoolSizeMax', can we also change those?

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/SocketObjectFactory.java
```diff
@@ -0,0 +1,91 @@
...
+    o.close();
+  }
+
+  @Override
+  public boolean validate(TSocket o)
+  {
+    boolean isClosed = (o != null && !o.isOpen());
+
+    // Saw that transport.close did not close socket, explicitly closing socket
+    if (isClosed && o != null) {
```
harmandeeps, line 65: I think we can remove 'o != null' ?

rubix-spi/src/test/java/com/qubole/rubix/spi/client/TestPoolingClient.java
```diff
@@ -0,0 +1,173 @@
...
+          e.printStackTrace();
+        }
+        assertTrue(message.get() == null, "Request should not have finished");
+        request.setSleepSecs(0);
+      }
+    }).start();
+
+    message.set(client.echo(request));
+
+    // given enough retries, we should have the response
```
harmandeeps, line 91: shouldn't we wait for the thread to finish?

rubix-tests/src/test/java/com/qubole/rubix/tests/TestNonLocalReadRequestChain.java
stagraqubole: Yes, changing

rubix-spi/src/main/java/com/qubole/rubix/spi/RetryingPoolingClient.java
stagraqubole: Because in borrowObject will remove it when it is picked later.

But I am changing this code to remove bad objects in `returnObject` itself. And doing it inside `returnObject` instead of letting users call some methods so that Pool's internal logic of good/bad object handling is not exposed unnecessarily to clients.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPool.java
stagraqubole: No, we are checking if there is atleast 1 connection which is open in the pool. `totalCount` in the ObjectPartitionPool keeps that count of total connections. As long as there is a connection active (may or may not be available right now to service request) we can expect that the request will be served

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
stagraqubole: changing it to `getTranportPoolMaxSize`

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/SocketObjectFactory.java
stagraqubole: Possible from first condition that isCLosed is true because o is null, so we still need this `o != null` here as we are going to dereference `o`

rubix-spi/src/test/java/com/qubole/rubix/spi/client/TestPoolingClient.java
stagraqubole: with the number of retries=20 and socketTimeout=500msec, the thread is going to finish before retries finish. In case it doesnt, the test will fail anyways so we dont need to wait.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -271,7 +271,7 @@ public static int getServerSocketTimeout(Configuration conf)
     return conf.getInt(KEY_SERVER_SOCKET_TIMEOUT, DEFAULT_SERVER_SOCKET_TIMEOUT);
   }
 
-  public static int getPoolSizeMax(Configuration conf)
+  public static int getTranportPoolMaxSize(Configuration conf)
```
harmandeeps, line 274: sorry, I forgot to mention: getPoolSizeMin, getPoolDeltaSize, getPoolMaxWait

stagraqubole: done

PR #356, opened 2020-04-15
Use positional read APIs for downloading file from cloud FS

raunaqmorarka:


stagraqubole:
With current block-size of 1MB, it is okay to do positional reads over List<ReadRequest> as two readRequests in the list will atleast be 1MB apart. But if we ever bring down the block size to much smaller value in future, then positional reads might cause slowness.

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -202,28 +203,17 @@ public int read(byte[] buffer, int offset, int length)
     }
     catch (Exception e) {
       log.error(String.format("Failed to read from rubix for file %s position %d length %d. Falling back to remote", remotePath, nextReadPosition, length), e);
-      getParentDataInputStream().seek(nextReadPosition);
-      int read = readFullyDirect(buffer, offset, length);
-      if (read > 0) {
-        nextReadPosition += read;
-        setNextReadBlock();
+      try {
+        getParentDataInputStream().readFully(nextReadPosition, buffer, offset, length);
```
stagraqubole, line 207: Isnt streaming read fallback better for streaming read() call? Ideally we should have the positional read override too here and in it's exception handling readFully is justified.

stagraqubole:
RemoteReadRequestChain can also use positional reads for all but first and the last ReadRequest

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
raunaqmorarka: Right, made changes to continue with streaming read fallback here and implemented positioned read API separately.

raunaqmorarka:
> RemoteReadRequestChain can also use positional reads for all but first and the last ReadRequest

Didn't get the part about first and last ReadRequest. When I tried to convert all reads in RemoteReadRequestChain to positioned reads, I got failures in TestCachingInputStream#testChunkCachingAndEviction

raunaqmorarka:
> With current block-size of 1MB, it is okay to do positional reads over List as two readRequests in the list will atleast be 1MB apart. But if we ever bring down the block size to much smaller value in future, then positional reads might cause slowness.

Right, I think it would take very small block size and lots of holes in the read requests list to cause slowdown. Suppose steaming reads do better in such case by reading additional data and discarding it, then we could simple stick to larger block size as that will at least cache the extra data downloaded.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloadRequestChain.java
```diff
@@ -156,39 +152,33 @@ public Integer call() throws IOException
...
-        int writeInThisCycle = Math.min(leftToWrite, directBuffer.capacity());
-        directBuffer.clear();
-        directBuffer.put(buffer, writtenSoFar, writeInThisCycle);
-        directBuffer.flip();
-        int nwrite = fileChannel.write(directBuffer, cacheReadStart + writtenSoFar);
-        directBuffer.compact();
-        writtenSoFar += nwrite;
-        leftToWrite -= nwrite;
+    while (nread < length) {
+      int nbytes = inputStream.read(cacheReadStart + nread, buffer, nread, length - nread);
```
stagraqubole, line 160: There seems to be a bug in existing code where we are allowing to read < length but later try to write to file `length` bytes. But we would not hit it given we already correct the ReadRequest backendReadEnd if it is going beyond eof.
We can replace this while loop with readFully call.

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -230,11 +234,39 @@ int readFullyDirect(byte[] buffer, int offset, int length)
...
+    catch (InterruptedException e) {
+      throw Throwables.propagate(e);
+    }
+    catch (Exception e) {
+      log.error(String.format("Failed to read from rubix for file %s position %d length %d. Falling back to remote", remotePath, nextReadPosition, length), e);
+      try {
+        getParentDataInputStream().readFully(position, buffer, offset, length);
+      }
+      catch (Exception ex) {
+        log.error(format("Error reading %d bytes directly from remote for file %s at position %d", length, remotePath, nextReadPosition), ex);
```
stagraqubole, line 255: wouldnt this exception be part of engine error/logs? We would do double logging with this

rubix-core/src/main/java/com/qubole/rubix/core/DummyModeCachingInputStream.java
```diff
@@ -55,18 +55,40 @@ public long getPos()
...
+  @Override
+  public int read(long position, byte[] buffer, int offset, int length)
+          throws IOException
+  {
+    long initPos = getPos();
+    try {
+      getParentDataInputStream().readFully(position, buffer, offset, length);
+    }
+    catch (Exception e) {
+      log.error(format("Error reading %d bytes directly from remote for file %s at position %d", length, remotePath, position), e);
```
stagraqubole, line 81: same here about the double log

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloadRequestChain.java
raunaqmorarka: Are we assured here that `length` won't go beyond EOF ? I had avoided readFully here because I wasn't sure whether `readRequest.getBackendReadLength()` takes into account size of file while aligning reads to block boundaries.
I think `int leftToWrite = nread;` would fix the bug you've mentioned ?

stagraqubole:
> > RemoteReadRequestChain can also use positional reads for all but first and the last ReadRequest
> 
> Didn't get the part about first and last ReadRequest. When I tried to convert all reads in RemoteReadRequestChain to positioned reads, I got failures in TestCachingInputStream#testChunkCachingAndEviction

First and lost blocks of RemoteRRC can contain prefix/suffix parts too which are created to align the read to block boundary. These parts would be < 1MB hence better to read first block with streaming read so that when we copy the prefix into prefixBuffer we might read ahead and get rest of the data for dataBuffer. Same for last block.

The tests should not fail due to this, this is an optimization and positional reads can still be used

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloadRequestChain.java
stagraqubole: Yes, that or using readFully (that will error out incase we failed to correct the end of read) should do

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
raunaqmorarka: ya, i've removed it now

rubix-core/src/main/java/com/qubole/rubix/core/DummyModeCachingInputStream.java
raunaqmorarka: Removed log

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloadRequestChain.java
raunaqmorarka: Used readFully here

raunaqmorarka:
> > > RemoteReadRequestChain can also use positional reads for all but first and the last ReadRequest
> > 
> > 
> > Didn't get the part about first and last ReadRequest. When I tried to convert all reads in RemoteReadRequestChain to positioned reads, I got failures in TestCachingInputStream#testChunkCachingAndEviction
> 
> First and lost blocks of RemoteRRC can contain prefix/suffix parts too which are created to align the read to block boundary. These parts would be < 1MB hence better to read first block with streaming read so that when we copy the prefix into prefixBuffer we might read ahead and get rest of the data for dataBuffer. Same for last block.
> 
> The tests should not fail due to this, this is an optimization and positional reads can still be used

Why are we reading the prefix and suffix parts separately in RemoteRRC ? We seem to be making 3 contiguous reads, can we make a single ranged read from `readRequest.backendReadStart` to `readRequest.backendReadEnd` here ?
Apart from RemoteRRC, rest of the changes are ready.

stagraqubole:
> > > > RemoteReadRequestChain can also use positional reads for all but first and the last ReadRequest
> > > 
> > > 
> > > Didn't get the part about first and last ReadRequest. When I tried to convert all reads in RemoteReadRequestChain to positioned reads, I got failures in TestCachingInputStream#testChunkCachingAndEviction
> > 
> > 
> > First and lost blocks of RemoteRRC can contain prefix/suffix parts too which are created to align the read to block boundary. These parts would be < 1MB hence better to read first block with streaming read so that when we copy the prefix into prefixBuffer we might read ahead and get rest of the data for dataBuffer. Same for last block.
> > The tests should not fail due to this, this is an optimization and positional reads can still be used
> 
> Why are we reading the prefix and suffix parts separately in RemoteRRC ? We seem to be making 3 contiguous reads, can we make a single ranged read from `readRequest.backendReadStart` to `readRequest.backendReadEnd` here ?
> Apart from RemoteRRC, rest of the changes are ready.

Because the buffers used are different, client buffer is sized as per actual read and cannot accommodate prefix/suffix.

PR #365, opened 2020-04-29
chg: dev: #364: Allow disabling data caching on master

### LINKED ISSUE #364: Add no-cache-on-master operation mode
Add mode of operation where caching is forcibly disabled on master node. In such mode Rubix will not validate presence of cache directories on master and all requests for data will be directly routed to source.
The need for that comes from https://github.com/prestosql/presto/issues/3488

losipiuk:
Fixes #364

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@
 
 # Maven
 target
+dependency-reduced-pom.xml
 
 # Sphinx
 _build
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
@@ -635,11 +635,6 @@ public abstract class BookKeeper implements BookKeeperService.Iface
         .build();
   }
 
-  public FileMetadata getEntry(String key, Callable<FileMetadata> callable) throws ExecutionException
-  {
-    return fileMetadataCache.get(key, callable);
-  }
-
   private static void initializeFileInfoCache(final Configuration conf, final Ticker ticker)
   {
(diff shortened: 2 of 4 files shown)


losipiuk:
@stagraqubole this is an attempt to allow configuring rubix in such a way that data caching is disabled on master node. Initialization does not fail then if data directories are missing. If configured such way I stub `fileMetadataCache` and `fileInfoCache` with implementation which disallow any modification. This serves as an assertion that we are not actually trying to cache any data on master.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/ImmutableEmptyCache.java
```diff
@@ -0,0 +1,108 @@
...
+  public V getIfPresent(Object key)
+  {
+    return null;
+  }
+
+  @Override
+  public V get(K key, Callable<? extends V> loader)
+          throws ExecutionException
+  {
+    return null;
```
stagraqubole, line 48: Throwing exception explicitly would make it easier to differentiate a case of response coming from valid Cache vs this ImmutableEmptyCache.

stagraqubole: There isn't. The cache objects are used only in the apis which can be invoked locally during read of a file. If we add something of that sorts in future then that will fail but it wont also be right to return empty result unless expected, so there too it is good to fail loudly if this mode is missed in those kind of changes.

losipiuk: Updated!

PR #366, opened 2020-04-30
Skip cache in case bookkeeper server is not up and cache directory is not created

harmandeeps:


rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -204,11 +204,19 @@ public FSDataInputStream open(Path path, int bufferSize)
...
+      return new FSDataInputStream(
+              new BufferedFSInputStream(
+                      new CachingInputStream(this, originalPath, this.getConf(), statsMBean,
+                              clusterManager.getClusterType(), bookKeeperFactory, fs, bufferSize, statistics),
+                      CacheConfig.getBlockSize(getConf())));
+    }
+    catch (Exception e) {
+      log.error("Error in opening Caching Input Stream, skipping cache", e);
+      inputStream = fs.open(path, bufferSize);
+      cacheSkipped = true;
```
rohangarg, line 217: 1. this is an FS object level variable - is it correct to set it here and not unset it after the bookkeeper server is up?
2. shouldn't we check whether the bookkeeper server is up or not in a separate thread and use cache only when it's up. else fallback to remote - instead of catching exception for all opening of input streams?

harmandeeps: 1. We already set FS object at line 190, in case of skip cache and this variable cacheSkipped is object variable, not a static, so next query run won't be affected.
2. We check 'fcache' directory which only gets created if bks is up while creating CachingInputStream. If not, we are falling back to the default file system. And, in my opinion, this is a simple solution.

rohangarg: 1. I see the `skipCache` util method takes path into input as well. so not sure if the previous occurrence is right too. further this is also being used in `getFileBlockLocations` - so not sure about the correctness of that occurrence too. something seems inconsistent to me.
2. aren't we creating a connection to bookkeeper server as well in the CachingInputStream's initialize? also shouldn't we try to reduce the overhead as much as possible for cases where the cache isn't even ready to use?

harmandeeps: 2. There is a method `initialize` where we initialize variables like local path (we check if fcache directory exists), and then we create a connection to BKS. Definitely, we need to reduce overhead, but we fall back as soon as we find any issue. If you keep a thread, you will be sending a heartbeat to BKS and keep it updated, so that solution has its own complexity and I don't see greater advantage in doing that.

stagraqubole: @harmandeeps lets remove the setting of `cacheSkipped` here. Another PR is getting rid of this variable.

stagraqubole: Also, in exception fail if CacheConfig.isStrictMode(conf) is set

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -199,11 +199,21 @@ public FSDataInputStream open(Path path, int bufferSize)
...
+                      new CachingInputStream(this, originalPath, this.getConf(), statsMBean,
+                              clusterManager.getClusterType(), bookKeeperFactory, fs, bufferSize, statistics),
+                      CacheConfig.getBlockSize(getConf())));
+    }
+    catch (Exception e) {
+      if (CacheConfig.isStrictMode(this.getConf())) {
+          log.error("Error in opening Caching Input Stream", e);
+          throw e;
+      }
+      log.error("Error in opening Caching Input Stream, skipping cache", e);
```
stagraqubole, line 214: This should be WARN log type

PR #374, opened 2020-05-05
Document the dependencies and build process

hashhar:
Add detailed steps so that people can more easily get set up with Rubix and contribute faster to Rubix.

### DIFF
--- a/docs/contrib/env.md
+++ b/docs/contrib/env.md
@@ -4,6 +4,25 @@ Rubix is a Maven project and uses Java 8. It uses JUnit as the testing framework
 Ensure that you have a development environment that support the above 
 configuration.
 
+## Pre-requisites
+
+* `thrift` binary needs to be available at `/usr/local/bin/thrift`.
+  For Debian-based systems:
+  ```sh
+  sudo apt-get install thrift-compiler
+  sudo ln -s /usr/bin/thrift /usr/local/bin/thrift
+  ```
+  For RPM-based systems:
+  ```sh
+  sudo yum install epel-release
+  sudo yum install thrift
+  sudo ln -s /usr/bin/thrift /usr/local/bin/thrift
+  ```
+* Java JDK 8 needs to be used since JDK versions > 9 don't ship with `tools.jar`. So if you see an error like `Could not find artifact jdk.tools:jdk.tools:jar:1.6 at specified path` then setup your system to use JDK 8 for the build.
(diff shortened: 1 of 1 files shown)


hashhar:
Unless people can build the project, they can't contribute code - let's make it easier for them to do.

docs/contrib/env.md
```diff
@@ -4,6 +4,25 @@ Rubix is a Maven project and uses Java 8. It uses JUnit as the testing framework
 Ensure that you have a development environment that support the above 
 configuration.
 
+## Pre-requisites
+
+* `thrift` binary needs to be available at `/usr/local/bin/thrift`.
+  For Debian-based systems:
+  ```sh
+  sudo apt-get install thrift-compiler
```
stagraqubole, line 12: From what I remember, latest version of thrift doesn't work until we upgrade and I needed to install 0.9.6 specifically. @raunaqmorarka did you also see issues with latest version of thrift binary?

raunaqmorarka: I did not try the latest version of thrift binary. Current docker image for build contains thrift version 0.9.3 built from source. On a dev box we could use any version of thrift but for production we would want the same version of thrift on the deployment and build machine.

hashhar: I got the build to work with 0.9.1 too. I think @raunaqmorarka is correct that the versions should ideally match on build and deploy.

stagraqubole: I tried this with thrift 0.11.0, which is installed with apt-get by default. Build fails with it, exactly like in this case: https://www.eclipse.org/lists/sw360-dev/msg00021.html

So version is important and we need to mention one. I also use 0.9.3, we can mention installation steps for that which you can find in [Dockerfile](https://github.com/qubole/rubix/blob/master/Dockerfile)

hashhar: Sure, I'll update this to point the exact version and add a note regarding trying to match build and deploy versions.

hashhar: I opted to point to upstream docs since adding the steps for both Debian and CentOS here made the doc too long and mean that the package names and steps would have needed effort to make sure they stay in sync.

I added a note about taking care about thrift versions and the version Qubole mostly uses. Feel free to reword if something looks off.

stagraqubole: Changed to mention the steps to build from source which should be common across systems.

stagraqubole:
Thanks for the contribution, merged it with few updates.

PR #378, opened 2020-05-07
Implement positioned reads in RemoteReadRequestChain

raunaqmorarka:


rubix-core/src/main/java/com/qubole/rubix/core/RemoteReadRequestChain.java
```diff
@@ -99,35 +99,48 @@ public Long call()
...
-        log.debug(String.format("Read %d bytes into destination buffer", readBytes));
-        copyIntoCache(fileChannel, directBuffer, readRequest.destBuffer, readRequest.destBufferOffset, readBytes, readRequest.actualReadStart);
-        log.debug(String.format("Copied %d requested bytes into cache", readBytes));
-        totalRequestedRead += readBytes;
+        int readBytes;
+        if (prefixBufferLength > 0 || suffixBufferLength > 0) {
+          // For single readRequest in chain, prefix and suffix both may be present
+          // seek needed in case of just suffix being present, otherwise it is no-op
+          inputStream.seek(readRequest.actualReadStart);
+          readBytes = readIntoBuffer(readRequest.getDestBuffer(), readRequest.destBufferOffset, readRequest.getActualReadLengthIntUnsafe(), inputStream);
```
stagraqubole, line 122: checkState(readBytes == readRequest.getActualReadLengthIntUnsafe())

raunaqmorarka: This check is currently there in `if (suffixBufferLength > 0)` block, I can move that here. Earlier we could see exception due to this in `Bookeeper#readDataInternal`, with the changes there we can add validation here

stagraqubole: Ya, lets move it here for readability. Also lets add checkState of each of the reads here: prefix, data and suffix as we are marking block in cache later even if we didnt read all data.

PR #381, opened 2020-05-11
add generation Number

suryanshagnihotri:
Tested the changes by making invalidation happen by : 
- Setting cache timeout  (KEY_DATA_CACHE_EXPIRY_AFTER_WRITE) to a K second and firing long-running query. Invalidation happens at regular interval and the query does not fail. (generation number increases)
- Setting the size of the cache  (KEY_MAX_CACHE_SIZE) to K MB and firing the query to scan data more than K MB from 2-3 tables. In this case, also invalidation happens and verified queries do not fail.
- Manually deleting the cache directory while a query was running, verified query does not fail and generation number increases.
grepped the log directory to check any errors too.

### DIFF
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
@@ -25,6 +25,8 @@ import com.google.common.cache.RemovalListener;
 import com.google.common.cache.RemovalNotification;
 import com.google.common.cache.Weigher;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.hash.BloomFilter;
+import com.google.common.hash.Funnels;
 import com.google.common.util.concurrent.Service;
 import com.qubole.rubix.bookkeeper.utils.DiskUtils;
 import com.qubole.rubix.bookkeeper.validation.CachingValidator;
@@ -40,8 +42,10 @@ import com.qubole.rubix.spi.ClusterType;
 import com.qubole.rubix.spi.thrift.BlockLocation;
 import com.qubole.rubix.spi.thrift.BookKeeperService;
 import com.qubole.rubix.spi.thrift.CacheStatusRequest;
+import com.qubole.rubix.spi.thrift.CacheStatusResponse;
 import com.qubole.rubix.spi.thrift.FileInfo;
 import com.qubole.rubix.spi.thrift.Location;
+import com.qubole.rubix.spi.thrift.ReadResponse;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
@@ -59,6 +63,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.nio.charset.Charset;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -83,6 +88,7 @@ import static com.qubole.rubix.common.metrics.BookKeeperMetrics.CacheMetric.CACH
 import static com.qubole.rubix.common.metrics.BookKeeperMetrics.CacheMetric.NONLOCAL_REQUEST_COUNT;
 import static com.qubole.rubix.common.metrics.BookKeeperMetrics.CacheMetric.REMOTE_REQUEST_COUNT;
 import static com.qubole.rubix.common.metrics.BookKeeperMetrics.CacheMetric.TOTAL_REQUEST_COUNT;
+import static com.qubole.rubix.spi.CacheUtil.UNKONWN_GENERATION_NUMBER;
 import static com.qubole.rubix.spi.ClusterType.TEST_CLUSTER_MANAGER;
 import static com.qubole.rubix.spi.ClusterType.TEST_CLUSTER_MANAGER_MULTINODE;
 import static com.qubole.rubix.spi.utils.DataSizeUnits.BYTES;
@@ -93,6 +99,8 @@ import static com.qubole.rubix.spi.utils.DataSizeUnits.MEGABYTES;
  */
 public abstract class BookKeeper implements BookKeeperService.Iface
 {
+  private static final int MAX_FILES_EXPECTED = 1000000;
+  private static final double FILE_ACCESSED_FILTER_FPP = 0.01;
   private static Log log = LogFactory.getLog(BookKeeper.class);
   private static DirectBufferPool bufferPool = new DirectBufferPool();
 
@@ -123,6 +131,10 @@ public abstract class BookKeeper implements BookKeeperService.Iface
   private Counter remoteRequestCount;
   private Counter cacheRequestCount;
   private Counter nonlocalRequestCount;
+
+  //  Maintains generation number for remote file
+  private Cache<String, Integer> generationNumberCache;
+  private BloomFilter fileAccessedBloomFilter;
 
   public BookKeeper(Configuration conf, BookKeeperMetrics bookKeeperMetrics) throws FileNotFoundException
   {
@@ -233,7 +245,7 @@ public abstract class BookKeeper implements BookKeeperService.Iface
   }
 
   @Override
-  public List<BlockLocation> getCacheStatus(CacheStatusRequest request) throws TException
+  public CacheStatusResponse getCacheStatus(CacheStatusRequest request) throws TException
   {
     try {
       initializeClusterManager(request);
@@ -275,11 +287,26 @@ public abstract class BookKeeper implements BookKeeperService.Iface
     }
 
     FileMetadata md;
+    //  If multiple threads call get of guava cache, then all but one get blocked and computation is done for it
+    //  and values is returned to other blocked threads.
     try {
-      md = fileMetadataCache.get(remotePath, new CreateFileMetadataCallable(remotePath, fileLength, lastModified, 0, conf));
+      md = fileMetadataCache.get(remotePath, new CreateFileMetadataCallable(
+              remotePath,
+              fileLength,
+              lastModified,
+              0,
+              conf,
+              generationNumberCache,
+              fileAccessedBloomFilter));
       if (isInvalidationRequired(md.getLastModified(), lastModified)) {
         invalidateFileMetadata(remotePath);
-        md = fileMetadataCache.get(remotePath, new CreateFileMetadataCallable(remotePath, fileLength, lastModified, 0, conf));
+        md = fileMetadataCache.get(remotePath, new CreateFileMetadataCallable(remotePath,
+                fileLength,
+                lastModified,
+                0,
+                conf,
+                generationNumberCache,
+                fileAccessedBloomFilter));
       }
     }
     catch (ExecutionException e) {
@@ -327,8 +354,7 @@ public abstract class BookKeeper implements BookKeeperService.Iface
       cacheRequestCount.inc(cacheRequests);
       remoteRequestCount.inc(remoteRequests);
     }
-
-    return blockLocations;
+    return new CacheStatusResponse(blockLocations, md.getGenerationNumber());
   }
 
   public boolean isInitialized()
@@ -423,16 +449,22 @@ public abstract class BookKeeper implements BookKeeperService.Iface
   }
 
   @Override
-  public void setAllCached(String remotePath, long fileLength, long lastModified, long startBlock, long endBlock)
+  public void setAllCached(String remotePath, long fileLength, long lastModified, long startBlock, long endBlock, int generationNumber)
       throws TException
   {
     FileMetadata md;
     md = fileMetadataCache.getIfPresent(remotePath);
 
     //md will be null when 2 users try to update the file in parallel and both their entries are invalidated.
+    // different generation number means invalidation has occurred
     // TODO: find a way to optimize this so that the file doesn't have to be read again in next request (new data is stored instead of invalidation)
     if (md == null) {
       log.warn(String.format("Could not update the metadata for file %s", remotePath));
+      return;
+    }
+    if (md.getGenerationNumber() != generationNumber)
+    {
+      log.debug(String.format("Could not update the metadata for file %s having different generationNumber %d in cache and %d in request", remotePath, md.getGenerationNumber(), generationNumber));
       return;
     }
     if (isInvalidationRequired(md.getLastModified(), lastModified)) {
@@ -498,14 +530,14 @@ public abstract class BookKeeper implements BookKeeperService.Iface
   // We cannot make it static because the variable is used in remoteReadRequestChain to write (cache) data to files.
   // So, it has to store the correct value in each thread. getCacheStatus is called twice.
   @Override
-  public boolean readData(String remotePath, long offset, int length, long fileSize, long lastModified, int clusterType)
+  public ReadResponse readData(String remotePath, long offset, int length, long fileSize, long lastModified, int clusterType)
       throws TException
   {
     if (CacheConfig.isParallelWarmupEnabled(conf)) {
       startRemoteFetchProcessor();
-      log.debug("Adding to the queue Path : " + remotePath + " Offste : " + offset + " Length " + length);
+      log.debug("Adding to the queue Path : " + remotePath + " Offset : " + offset + " Length " + length);
       fetchProcessor.addToProcessQueue(remotePath, offset, length, fileSize, lastModified);
-      return true;
+      return new ReadResponse(false, UNKONWN_GENERATION_NUMBER);
     }
     else {
       return readDataInternal(remotePath, offset, length, fileSize, lastModified, clusterType);
@@ -531,21 +563,26 @@ public abstract class BookKeeper implements BookKeeperService.Iface
     }
   }
 
-  private boolean readDataInternal(String remotePath, long offset, int length, long fileSize,
-                                   long lastModified, int clusterType) throws TException
+  private ReadResponse readDataInternal(String remotePath,
+      long offset,
+      int length,
+      long fileSize,
+      long lastModified,
+      int clusterType) throws TException
   {
     int blockSize = CacheConfig.getBlockSize(conf);
     byte[] buffer = new byte[blockSize];
-    String localPath = CacheUtil.getLocalPath(remotePath, conf);
     FileSystem fs = null;
     FSDataInputStream inputStream = null;
     Path path = new Path(remotePath);
     long startBlock = offset / blockSize;
     long endBlock = ((offset + (length - 1)) / CacheConfig.getBlockSize(conf)) + 1;
+    CacheStatusResponse response = null;
     try {
       int idx = 0;
       CacheStatusRequest request = new CacheStatusRequest(remotePath, fileSize, lastModified, startBlock, endBlock).setClusterType(clusterType);
-      List<BlockLocation> blockLocations = getCacheStatus(request);
+      response = getCacheStatus(request);
+      List<BlockLocation> blockLocations = response.getBlocks();
 
       for (long blockNum = startBlock; blockNum < endBlock; blockNum++, idx++) {
         long readStart = blockNum * blockSize;
@@ -562,7 +599,7 @@ public abstract class BookKeeper implements BookKeeperService.Iface
           // Ue RRRC directly instead of creating instance of CachingFS as in certain circumstances, CachingFS could
           // send this request to NonLocalRRC which would be wrong as that would not cache it on disk
           long expectedBytesToRead = (readStart + blockSize) > fileSize ? (fileSize - readStart) : blockSize;
-          RemoteReadRequestChain remoteReadRequestChain = new RemoteReadRequestChain(inputStream, localPath, bufferPool, CacheConfig.getDiskReadBufferSize(conf), buffer, new BookKeeperFactory(this));
+          RemoteReadRequestChain remoteReadRequestChain = new RemoteReadRequestChain(inputStream, remotePath, response.getGenerationNumber(), bufferPool, conf, buffer, new BookKeeperFactory(this));
           remoteReadRequestChain.addReadRequest(new ReadRequest(readStart, readStart + expectedBytesToRead, readStart, readStart + expectedBytesToRead, buffer, 0, fileSize));
           remoteReadRequestChain.lock();
           long dataRead = remoteReadRequestChain.call();
@@ -574,15 +611,15 @@ public abstract class BookKeeper implements BookKeeperService.Iface
           }
           else {
             log.error("Not able to download requested bytes. Not updating the cache for block " + blockNum);
-            return false;
+            return new ReadResponse(false, response.getGenerationNumber());
           }
         }
       }
-      return true;
+      return new ReadResponse(true, response.getGenerationNumber());
     }
     catch (Exception e) {
       log.warn("Could not cache data: ", e);
-      return false;
+      return new ReadResponse(false, response.getGenerationNumber());
     }
     finally {
       if (inputStream != null) {
@@ -606,7 +643,7 @@ public abstract class BookKeeper implements BookKeeperService.Iface
     return endBlock;
   }
 
-  private static synchronized void initializeCache(final Configuration conf, final Ticker ticker)
+  private synchronized void initializeCache(final Configuration conf, final Ticker ticker)
       throws FileNotFoundException
   {
     if (CacheConfig.isOnMaster(conf) && !CacheConfig.isCacheDataOnMasterEnabled(conf)) {
@@ -665,6 +702,11 @@ public abstract class BookKeeper implements BookKeeperService.Iface
         .expireAfterWrite(CacheConfig.getCacheDataExpirationAfterWrite(conf), TimeUnit.MILLISECONDS)
         .removalListener(new CacheRemovalListener())
         .build();
+    fileAccessedBloomFilter = BloomFilter.create(Funnels.stringFunnel(Charset.defaultCharset()), MAX_FILES_EXPECTED, FILE_ACCESSED_FILTER_FPP);
+    generationNumberCache = CacheBuilder.newBuilder()
+            .expireAfterAccess(2, TimeUnit.HOURS)
+            .build();
+
   }
 
   @VisibleForTesting
@@ -740,21 +782,30 @@ public abstract class BookKeeper implements BookKeeperService.Iface
     long fileLength;
     long lastModified;
     long currentFileSize;
+    Cache<String, Integer> generationNumberCache;
+    BloomFilter fileAccessedBloomFilter;
 
-    public CreateFileMetadataCallable(String path, long fileLength, long lastModified, long currentFileSize,
-                                      Configuration conf)
+    public CreateFileMetadataCallable(String path,
+        long fileLength,
+        long lastModified,
+        long currentFileSize,
+        Configuration conf,
+        Cache<String, Integer> generationNumberCache,
+        BloomFilter fileAccessedBloomFilter)
     {
       this.path = path;
       this.conf = conf;
       this.fileLength = fileLength;
       this.lastModified = lastModified;
       this.currentFileSize = currentFileSize;
+      this.generationNumberCache = generationNumberCache;
+      this.fileAccessedBloomFilter = fileAccessedBloomFilter;
     }
 
     public FileMetadata call()
         throws Exception
     {
-      return new FileMetadata(path, fileLength, lastModified, currentFileSize, conf);
+      return new FileMetadata(path, fileLength, lastModified, currentFileSize, conf, generationNumberCache, fileAccessedBloomFilter);
     }
   }
 
@@ -772,13 +823,18 @@ public abstract class BookKeeper implements BookKeeperService.Iface
     }
   }
 
-  private void replaceFileMetadata(String key, long currentFileSize, Configuration conf) throws IOException
+  private void replaceFileMetadata(String key, long currentFileSize, Configuration conf)
+      throws IOException
   {
     if (fileMetadataCache != null) {
       FileMetadata metadata = fileMetadataCache.getIfPresent(key);
       if (metadata != null) {
-        FileMetadata newMetaData = new FileMetadata(key, metadata.getFileSize(), metadata.getLastModified(),
-            currentFileSize, conf);
+        FileMetadata newMetaData = new FileMetadata(key,
+            metadata.getFileSize(),
+            metadata.getLastModified(),
+            currentFileSize,
+            conf,
+            metadata.getGenerationNumber());
         fileMetadataCache.put(key, newMetaData);
       }
     }
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloader.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloader.java
@@ -24,6 +24,7 @@ import com.qubole.rubix.spi.CacheConfig;
 import com.qubole.rubix.spi.CacheUtil;
 import com.qubole.rubix.spi.thrift.BlockLocation;
 import com.qubole.rubix.spi.thrift.CacheStatusRequest;
+import com.qubole.rubix.spi.thrift.CacheStatusResponse;
 import com.qubole.rubix.spi.thrift.Location;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -46,6 +47,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ThreadPoolExecutor;
 
 import static com.qubole.rubix.spi.utils.DataSizeUnits.BYTES;
+import static com.qubole.rubix.spi.CacheUtil.UNKONWN_GENERATION_NUMBER;
 import static com.qubole.rubix.spi.CommonUtilities.toBlockStartPosition;
 import static com.qubole.rubix.spi.CommonUtilities.toEndBlock;
 import static com.qubole.rubix.spi.CommonUtilities.toStartBlock;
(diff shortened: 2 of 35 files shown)


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -55,15 +58,29 @@ public FileMetadata()
...
     this.remotePath = remotePath;
     this.size = fileLength;
     this.lastModified = lastModified;
     this.currentFileSize = currentFileSize;
-    localPath = CacheUtil.getLocalPath(remotePath, conf);
-    mdFilePath = CacheUtil.getMetadataFilePath(remotePath, conf);
+    if (!updateMetadata) {
+      if (generationNumber.containsKey(remotePath)) {
+        genNumber = generationNumber.get(remotePath);
+        genNumber++;
```
stagraqubole, line 71: This block is not thread safe. Multiple getCacheStatus on file  can come in parallel. 
Make sure when you make this thread safe, you dont block the requests on other file on the exists() check later.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -123,6 +125,9 @@
   private Counter cacheRequestCount;
   private Counter nonlocalRequestCount;
 
+  // maintain generation number for remote file
+  protected static ConcurrentHashMap<String, Integer> generationNumber = new ConcurrentHashMap<>();
```
stagraqubole, line 129: memory is leaked on this map. This can cause big problems in long lived deployments that access millions of files over time

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -55,15 +58,29 @@ public FileMetadata()
...
+        while (Files.exists(Paths.get(CacheUtil.getMetadataFilePath(remotePath, conf, genNumber)))) {
+          genNumber++;
```
stagraqubole, line 73: You need to also consider the mode where it is expected to resume from the state on disk if rubix servers had crashed. This is tricky because we want all three things:
1. No memory leaks on generationNumber map
2. Support resuming from on-disk state
3. Also skip generationNumbers for which files already exist when the case is not of resuming from on-disk state

stagraqubole:
Couple of more things:
1. setAllCached api also needs to be made generationNumber aware because it is possible that it is called after caching data for old gen data file which will cause corruption as currently you are updating the cache status. 
2. We need a test that runs CachedReads in parallel to invalidations being forced

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -650,6 +663,12 @@ public int weigh(String key, FileMetadata md)
         .expireAfterWrite(CacheConfig.getCacheDataExpirationAfterWrite(conf), TimeUnit.MILLISECONDS)
         .removalListener(new CacheRemovalListener())
         .build();
+    log.warn("Initializing everything again");
```
stagraqubole, line 666: This log is not correct, remove

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -650,6 +663,12 @@ public int weigh(String key, FileMetadata md)
...
+    filter = BloomFilter.create(Funnels.stringFunnel(Charset.defaultCharset()), 1000000, 0.01);
```
stagraqubole, line 667: Define constants for 1000000 and 0.01 with appropriate names

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -650,6 +663,12 @@ public int weigh(String key, FileMetadata md)
...
+    generationNumber = CacheBuilder.newBuilder()
+            .expireAfterWrite(2, TimeUnit.HOURS)
```
stagraqubole, line 669: `expireAfterWrite` -> `expireAfterAccess`

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -41,30 +48,68 @@
...
+  private int generation;
 
   int bitmapFileSizeBytes;
   ByteBufferBitmap blockBitmap;
 
   static Striped<Lock> stripes = Striped.lock(20000);
+  //  Maintains generation number for remote file
+  //  If multiple threads call get of guava cache , then all are blocked and computation is done for one
+  //  and values is returned to other blocked threads.
+  static Cache<String, Integer> generationNumber = CacheBuilder.newBuilder()
```
stagraqubole, line 61: These are initialized in BookKeeper. Do this:
1. Move these declarations into BookKeeper
2. Drop static
3. Drop default definitions
4. Make BookKeeper.initializeCache as non-static so that you can define these non static objects in it.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -41,30 +48,68 @@
...
+          .expireAfterWrite(2, TimeUnit.HOURS)
+          .build();
+  static BloomFilter filter = BloomFilter.create(Funnels.stringFunnel(Charset.defaultCharset()), 1000000, 0.01);;
 
   private static Log log = LogFactory.getLog(FileMetadata.class.getName());
 
   public FileMetadata()
```
stagraqubole, line 68: Default constructor looks unused, you can drop it

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
```diff
@@ -152,12 +152,13 @@ public static String getDirPath(int dirIndex, Configuration conf)
...
    * @param conf        The current Hadoop configuration.
+   * @param generationNumber the generation number for local file corresponding to remote location
    * @return The local path location.
    */
-  public static String getLocalPath(String remotePath, Configuration conf)
+  public static String getLocalPath(String remotePath, Configuration conf, int generationNumber)
   {
     final String absLocation = getDirectory(remotePath, conf);
-    return absLocation + "/" + getName(remotePath);
+    return absLocation + "/" + getName(remotePath) + "g" + generationNumber;
```
harmandeeps, line 161: use string.format for better readability.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
```diff
@@ -173,12 +174,13 @@ public static String getRemotePath(String localPath, Configuration conf)
...
    * @param conf        The current Hadoop configuration.
+   * @param generationNumber the generation number for local file corresponding to remote location
    * @return The metadata file path.
    */
-  public static String getMetadataFilePath(String remotePath, Configuration conf)
+  public static String getMetadataFilePath(String remotePath, Configuration conf, int generationNumber)
   {
     final String absLocation = getDirectory(remotePath, conf);
-    return absLocation + "/" + getName(remotePath) + CacheConfig.getCacheMetadataFileSuffix(conf);
+    return absLocation + "/" + getName(remotePath) + CacheConfig.getCacheMetadataFileSuffix(conf) + "g" + generationNumber;
```
harmandeeps, line 183: same.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -233,8 +237,10 @@ public void run()
         log.debug(String.format("Trying to read from %s at offset %d and length %d for client %s", remotePath, offset, readLength, localDataTransferClient.getRemoteAddress()));
         try (RetryingPooledBookkeeperClient bookKeeperClient = bookKeeperFactory.createBookKeeperClient(conf)) {
           if (!CacheConfig.isParallelWarmupEnabled(conf)) {
-            if (!bookKeeperClient.readData(remotePath, offset, readLength, header.getFileSize(),
-                    header.getLastModified(), header.getClusterType())) {
+            ReadResponse response = bookKeeperClient.readData(remotePath, offset, readLength, header.getFileSize(),
+                    header.getLastModified(), header.getClusterType());
+            generationNumber = response.getGenerationNumber();
```
harmandeeps, line 242: put 'generationNumber = response.getGenerationNumber();' in else block.

harmandeeps: it is already there, so we can drop this.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloader.java
```diff
@@ -136,13 +131,24 @@ private void initializeMetrics()
...
                           startBlock,
                           endBlock));
+          blockLocations = response.getBlocks();
+          genNumber = response.getGenerationNumber();
+          localPath = CacheUtil.getLocalPath(remotePath, conf, genNumber);
+          log.debug("Processing Request for File : " + path.toString() + " LocalFile : " + localPath);
+
+          if (requestChain == null)
+          {
+            requestChain = new FileDownloadRequestChain(bookKeeper, fs, localPath,
```
harmandeeps, line 148: add debug logs here

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloader.java
```diff
@@ -109,13 +107,10 @@ private void initializeMetrics()
       FileSystem fs = FileSystem.get(path.toUri(), conf);
       fs.initialize(path.toUri(), conf);
 
-      String localPath = CacheUtil.getLocalPath(entry.getKey(), conf);
-      log.debug("Processing Request for File : " + path.toString() + " LocalFile : " + localPath);
+      String localPath;
```
harmandeeps, line 110: seems like this is not used outside try block, better declare it there only.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -41,30 +48,68 @@
...
     this.remotePath = remotePath;
     this.size = fileLength;
     this.lastModified = lastModified;
     this.currentFileSize = currentFileSize;
-    localPath = CacheUtil.getLocalPath(remotePath, conf);
-    mdFilePath = CacheUtil.getMetadataFilePath(remotePath, conf);
-
+    Lock lock = stripes.get(remotePath);
+    try {
+      lock.lock();
```
harmandeeps, line 81: Please add a comment to briefly explain the logic and things like why the lock is needed etc.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/validation/CachingValidator.java
```diff
@@ -117,7 +119,11 @@ protected boolean validateCachingBehavior()
         fileLastModified, VALIDATOR_START_BLOCK, VALIDATOR_END_BLOCK).setClusterType(VALIDATOR_CLUSTER_TYPE);
 
     try {
-      List<BlockLocation> locations = bookKeeper.getCacheStatus(request);
+      CacheStatusResponse response = bookKeeper.getCacheStatus(request);
+      if (response == null) {
```
harmandeeps, line 123: this can be clubbed into one condition as:
```
if (response == null || response.getBlocks().isEmpty() || response.getBlocks().get(0).getLocation() != Location.LOCAL) {
        return false;	        return false;
}
```

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/validation/CachingValidator.java
```diff
@@ -128,12 +134,12 @@ protected boolean validateCachingBehavior()
...
           fileLength,
           fileLastModified,
-          VALIDATOR_CLUSTER_TYPE);
+          VALIDATOR_CLUSTER_TYPE).isStatus();
       if (!dataRead) {
         return false;
       }
 
-      locations = bookKeeper.getCacheStatus(request);
+      locations = bookKeeper.getCacheStatus(request).getBlocks();
```
harmandeeps, line 142: why do we need to check this again? I am missing something?

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheUtil.java
stagraqubole: And add a "_" before "g"

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -41,30 +48,68 @@
   private long size;
   private long lastModified;
   private long currentFileSize;
-
   private boolean needsRefresh = true;
+  private int generation;
```
harmandeeps, line 52: rename it to `generationNumber` as we have used this term everywhere for consistency.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -41,30 +48,68 @@
...
+        }
+        else {
+            genNumber = generationNumber.get(remotePath, new Callable<Integer>() {
+              @Override
+              public Integer call() throws Exception {
+                return -1;
+              }
+            });
+            genNumber++;
+            while (new File(CacheUtil.getLocalPath(remotePath, conf, genNumber)).exists()) {
```
harmandeeps, line 99: why is this logic needed if we are reading from cache? Also, cache may return '-1' if it doesn't have the path, we should add a check.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -41,30 +48,68 @@
...
-    mdFilePath = CacheUtil.getMetadataFilePath(remotePath, conf);
-
+    Lock lock = stripes.get(remotePath);
+    try {
+      lock.lock();
+      if (newFileEntry) {
+        int genNumber = 0;
+        if (!filter.mightContain(remotePath)) {
+          // BKS restarted
+          while (new File(CacheUtil.getLocalPath(remotePath, conf, genNumber)).exists()) {
```
harmandeeps, line 86: we can use 'listFiles(Filter)' (https://docs.oracle.com/javase/6/docs/api/java/io/File.html#list(java.io.FilenameFilter)) to list the files, instead of doing this in a sequential manner.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
harmandeeps: It can be initialized to '-1' to denote an invalid generation instead of. taking 0 as default.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
stagraqubole: This is just a failsafe, in case there are files existing already. This ideally should never happen

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/validation/CachingValidator.java
suryanshagnihotri: readData is called before it.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloader.java
```diff
@@ -65,6 +62,7 @@
   private MetricRegistry metrics;
   private Counter totalMBDownloaded;
   private Counter totalTimeToDownload;
+  private int generationNumber;
```
stagraqubole, line 65: You should not do this. FileDownloader is singleton in a system and you are keeping generationNumber for each instance here. This is also wrong as per the current logic where you keep the generationNumber of last file in the downloader and use it for all in setCacheStatus

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -206,6 +209,7 @@ public void stop()
     SocketChannel localDataTransferClient;
     Configuration conf;
     BookKeeperFactory bookKeeperFactory;
+    int generationNumber;
```
stagraqubole, line 212: Avoid adding this as class member. There is work going on to make this thread long lived and handle multiple requests. In that case, it would be cleaner to have this as a local variable

rubix-core/src/main/java/com/qubole/rubix/core/CachedReadRequestChain.java
```diff
@@ -49,11 +49,13 @@
   private DirectBufferPool bufferPool;
   private int directBufferSize;
   private int corruptedFileCount;
+  private int generationNumber;
```
stagraqubole, line 52: There are multiple places where generationNumber is getting initialized with the default value of int ie 0. I suggest we make '0' as the special case of UNKNOWN_GENERATION_NUMBER and assert in `getLocalPath` and `getMetadataFilePath` to not get this gen number. So valid gen numbers will start from 1 and we can easily catch issues of invalid generationNumber being used

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeper.java
```diff
@@ -478,19 +485,86 @@ public void testSetAllCacheIngoresCachedBlocks() throws Exception
...
+    Thread invalidateRequest = new Thread(new Runnable() {
+      @Override
+      public void run() {
+        for (int i = 1; i < 15; i++)
+        {
+          bookKeeper.invalidateFileMetadata(remotePathWithScheme);
+          try {
+            Thread.sleep(1000);
+          } catch (InterruptedException e) {
+            e.printStackTrace();
```
stagraqubole, line 529: throw this as runtimeException

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeper.java
```diff
@@ -478,19 +485,86 @@ public void testSetAllCacheIngoresCachedBlocks() throws Exception
...
+        for (int i = 1; i < 200; i++)
+        {
+          try {
+            ReadResponse response = bookKeeper.readData(remotePathWithScheme, readOffset, readLength, TEST_FILE_LENGTH, TEST_LAST_MODIFIED, ClusterType.TEST_CLUSTER_MANAGER.ordinal());
+            Thread.sleep(100);
+            assertEquals(response.isStatus(), true);
+          }
+          catch (Exception e)
+          {
+            fail();
```
stagraqubole, line 546: Same as above, avoid fail

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeper.java
```diff
@@ -478,19 +485,86 @@ public void testSetAllCacheIngoresCachedBlocks() throws Exception
...
     // Now a overlapping request of 10 new blocks, new size of FileMetadata should only increase for these 10 new blocks
-    bookKeeper.setAllCached(TEST_REMOTE_PATH, TEST_FILE_LENGTH, TEST_LAST_MODIFIED, TEST_START_BLOCK - 5, TEST_END_BLOCK + 5);
+    bookKeeper.setAllCached(TEST_REMOTE_PATH, TEST_FILE_LENGTH, TEST_LAST_MODIFIED, TEST_START_BLOCK - 5, TEST_END_BLOCK + 5, genNumber);
     long newSize = bookKeeper.getFileMetadata(TEST_REMOTE_PATH).getCurrentFileSize();
     assertTrue(newSize == size + 10 * CacheConfig.getBlockSize(conf),
             String.format("Expected size: %s but found %s", (size + 10) * CacheConfig.getBlockSize(conf), newSize));
   }
+
+  @Test
+  void testReadWithInvalidation() throws IOException, TException, InterruptedException {
```
stagraqubole, line 506: Instead of here, we should add this test in rubix-tests so that we get integration level test. In there we should do it in two parts:
1. Parallel warmup disabled
2. Parallel warmup enabled

In both cases, we need to read via CachingFileSystem while invalidating in parallel.

rubix-bookkeeper/src/test/java/com/qubole/rubix/bookkeeper/TestBookKeeper.java
```diff
@@ -478,19 +485,86 @@ public void testSetAllCacheIngoresCachedBlocks() throws Exception
...
+  {
+    final String remotePathWithScheme = "file://" + TEST_REMOTE_PATH;
+    final int readOffset = 0;
+    final int readLength = 2000;
+    DataGen.populateFile(TEST_REMOTE_PATH);
+    CacheStatusRequest request = new CacheStatusRequest(remotePathWithScheme, TEST_FILE_LENGTH, TEST_LAST_MODIFIED,
+            0, 20).setClusterType(ClusterType.TEST_CLUSTER_MANAGER.ordinal());
+     bookKeeper.readData(remotePathWithScheme, readOffset, readLength, TEST_FILE_LENGTH, TEST_LAST_MODIFIED, ClusterType.TEST_CLUSTER_MANAGER.ordinal());
+     bookKeeper.invalidateFileMetadata(remotePathWithScheme);
+     assertEquals(bookKeeper.getCacheStatus(request).getGenerationNumber(), 1);
```
stagraqubole, line 568: This will change after you bring in my commit to assert = 2

stagraqubole: Better change this to assert = (UNKNOWN_GENERATION_NUMBER + 2)

rubix-spi/src/test/java/com/qubole/rubix/spi/TestCacheUtil.java
```diff
@@ -153,8 +153,8 @@ public void testGetLocalPath()
 
     createCacheDirectoriesForTest(conf);
 
-    String localPath = CacheUtil.getLocalPath(remotePath, conf);
-    assertEquals(localPath, cacheTestDirPrefix + "0" + "/fcache/" + localRelPath, "Paths not equal!");
+    String localPath = CacheUtil.getLocalPath(remotePath, conf, 0);
+    assertEquals(localPath, cacheTestDirPrefix + "0" + "/fcache/" + localRelPath + "_g0", "Paths not equal!");
```
stagraqubole, line 157: Write a method here that returns expectedPath and use for all tests in this file. Make that method use UNKNOWN_GENERATION_NUMBER based prefix instead of hardcoding `_g1`

rubix-tests/src/test/java/com/qubole/rubix/tests/TestGenerationNumber.java
```diff
@@ -0,0 +1,279 @@
...
+    Thread readrequest = new Thread(new Runnable() {
+      @Override
+      public void run() {
+        for (int i = 1; i < 200; i++)
+        {
+          try {
+            byte[] buffer = new byte[readLength];
+            inputStream.seek(0);
+            int readSize = inputStream.read(buffer, 0, readLength);
+            assertTrue(readSize == readLength, "Wrong amount of data read " + readSize + " was expecting " + readLength);
```
stagraqubole, line 158: validate data read too, just the size read check is not enough

rubix-tests/src/test/java/com/qubole/rubix/tests/TestGenerationNumber.java
```diff
@@ -0,0 +1,279 @@
...
+          try {
+            client.invalidateFileMetadata(backendPath.toString());
+            Thread.sleep(1000);
+          } catch (Exception e) {
+            throw new RuntimeException(e);
+          }
+        }
+      }
+    });
+    Thread readrequest = new Thread(new Runnable() {
```
stagraqubole, line 149: Start multiple reader threads, each reading different range of file and in each cycle validate that right data is read

rubix-tests/src/test/java/com/qubole/rubix/tests/TestGenerationNumber.java
```diff
@@ -0,0 +1,279 @@
...
+
+  @Test
+  void testReadWithInvalidations() throws Exception {
+    testReadWithInvalidationHelper();
+    CacheConfig.setIsParallelWarmupEnabled(conf, true);
+    restartServer();
+    testReadWithInvalidationHelper();
+  }
+
+  void testReadWithInvalidationHelper() throws TException, InterruptedException {
```
stagraqubole, line 132: This tests local reads, we need to add similar test for non-local read as well.

rubix-tests/src/test/java/com/qubole/rubix/tests/TestGenerationNumber.java
```diff
@@ -0,0 +1,279 @@
...
+    return new CachingInputStream(backendPath, conf,
+        new CachingFileSystemStats(), ClusterType.TEST_CLUSTER_MANAGER,
+        new BookKeeperFactory(), localFileSystem,
+        CacheConfig.getBlockSize(conf), null);
+  }
+
+  @Test
+  void testReadWithInvalidations() throws Exception {
+    testReadWithInvalidationHelper();
+    CacheConfig.setIsParallelWarmupEnabled(conf, true);
```
stagraqubole, line 127: Just setting `CacheConfig.setIsParallelWarmupEnabled(conf, true)` won't be enough. Set the configs for `RemoteFetchProcessor` to be aggressive to start downloading files quickly (by default it will wait for 10sec)

rubix-tests/src/test/java/com/qubole/rubix/tests/TestGenerationNumber.java
```diff
@@ -0,0 +1,279 @@
...
+    restartServer();
+    testReadWithInvalidationHelper();
```
stagraqubole, line 129: Also have two different test methods for parallel warmup and normal case. Makes it easier to debug

rubix-tests/src/test/java/com/qubole/rubix/tests/TestGenerationNumber.java
```diff
@@ -0,0 +1,279 @@
...
+  {
+    int generationNumber = 5;
+    // create files till generation Number = 5
+    creatLocalFilesOnCache(generationNumber);
+    boolean missingFile = false;
+    testGenerationNumberDuringStartUpHelper(generationNumber, missingFile);
+    missingFile = true;
+    restartServer();
+    creatLocalFilesOnCache(generationNumber);
+    new File(CacheUtil.getMetadataFilePath(backendPath.toString(), conf, generationNumber)).delete();
```
stagraqubole, line 200: There are some cases still left. Cover all the cases covered in FileMetadata constructor to ensure no regressions in future.

rubix-tests/src/test/java/com/qubole/rubix/tests/TestGenerationNumber.java
```diff
@@ -0,0 +1,550 @@
...
+      this.inputStream = inputStream;
+    }
+    @Override
+    public Data call()
+    {
+      try {
+        byte[] buffer = new byte[readLength];
+        inputStream.seek(0);
+        int readSize = inputStream.read(buffer, 0, readLength);
+        Thread.sleep(2000);
```
stagraqubole, line 221: Why this sleep?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -99,6 +101,22 @@ public FileMetadata(String remotePath,
      */
   }
 
+  private void createLocalFiles()
```
stagraqubole, line 104: add comment to method signature that BKS is responsible for creating new files (data + mdfiles) via FileMetadata in all cases. RRCs should never create new files, if they ever hit a case of FNF then they should fail the request as it points to invalidations.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -52,20 +58,40 @@
...
+      BloomFilter fileAccessedBloomFilter)
+      throws ExecutionException, IOException
   {
+    this(remotePath,
+        fileLength,
+        lastModified,
+        currentFileSize,
+        conf,
+        findGenerationNumber(remotePath, conf, generationNumberCache, fileAccessedBloomFilter));
+    createLocalFiles();
```
stagraqubole, line 77: call createLocalFiles in the next constructor

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
stagraqubole: I see you have it already. Reword it so that the message is strong that this responsibility shouldnt change and move the comment before method name

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -423,18 +449,24 @@ public ClusterManager getClusterManagerInstance(ClusterType clusterType, Configu
...
     //md will be null when 2 users try to update the file in parallel and both their entries are invalidated.
+    // different generation number means invalidation has occurred
     // TODO: find a way to optimize this so that the file doesn't have to be read again in next request (new data is stored instead of invalidation)
     if (md == null) {
       log.warn(String.format("Could not update the metadata for file %s", remotePath));
       return;
     }
+    if (md.getGenerationNumber() != generationNumber)
+    {
+      log.warn(String.format("Could not update the metadata for file %s having different generationNumber %d in cache and %d in request", remotePath, md.getGenerationNumber(), generationNumber));
```
stagraqubole, line 467: mark the log as debug level. We dont need to bring this to focus as this is a valid case

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloader.java
```diff
@@ -45,11 +46,13 @@
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadPoolExecutor;
 
+import static com.qubole.rubix.spi.CacheUtil.UNKONWN_GENERATION_NUMBER;
 import static com.qubole.rubix.spi.utils.DataSizeUnits.BYTES;
 import static com.qubole.rubix.spi.CommonUtilities.toBlockStartPosition;
 import static com.qubole.rubix.spi.CommonUtilities.toEndBlock;
 import static com.qubole.rubix.spi.CommonUtilities.toStartBlock;
 
+
```
stagraqubole, line 55: remove extra line

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileDownloader.java
```diff
@@ -130,13 +130,31 @@ private void initializeMetrics()
...
+          CacheStatusResponse response = bookKeeper.getCacheStatus(
                   new CacheStatusRequest(
                           context.getRemoteFilePath(),
                           context.getFileSize(),
                           context.getLastModifiedTime(),
                           startBlock,
                           endBlock));
+          blockLocations = response.getBlocks();
+          if(generationNumber != UNKONWN_GENERATION_NUMBER && response.getGenerationNumber() != generationNumber) {
+            log.warn(String.format("Mismatch in generation-number in download requests for file %s, expected=%d but found=%d, skipping the file",
```
stagraqubole, line 142: This can also be moved to debug level

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -221,6 +224,7 @@ public void run()
         log.debug("Connected to node - Local Address: " + localDataTransferClient.getLocalAddress() +
                 " Remote Address: " + localDataTransferClient.getRemoteAddress());
 
+
```
stagraqubole, line 227: remove extra line

rubix-core/src/main/java/com/qubole/rubix/core/CachedReadRequestChain.java
```diff
@@ -49,31 +50,41 @@
...
```
stagraqubole, line 53: We dont need to keep another copy of `generationNumber`. Remove this and make `ReadRequestChain.generationNumber` scope `protected`

rubix-core/src/main/java/com/qubole/rubix/core/RemoteReadRequestChain.java
```diff
@@ -50,26 +52,35 @@
   private int blockSize;
 
   private BookKeeperFactory bookKeeperFactory;
+  private final int generationNumber;
```
stagraqubole, line 55: same here, we can remove `generationNumber` here

Issue #349, opened 2020-04-13
Bypass getFileInfo network call when staleness check is enabled


stagraqubole:
By default staleness check is enabled and in that case BookKeeper will make a fs.getFileStatus call to create FileInfo. This call could have been made in CachingInputStream instead when staleness check is enabled, it would save one bookkeeper network call per CachingInputStream

raunaqmorarka:
Fixed by https://github.com/qubole/rubix/pull/382

PR #382, opened 2020-05-14
Bypass getFileInfo network call when staleness check is enabled

raunaqmorarka:
Refactored test code to remove LocalFSInputStream and use RawLocalFileSystem
Simplified CachingInputStream to have single constructor which is also tested in UTs


raunaqmorarka:
Fixes https://github.com/qubole/rubix/issues/349

rubix-core/src/main/java/com/qubole/rubix/core/CachingInputStream.java
```diff
@@ -90,67 +89,48 @@ public Thread newThread(Runnable r)
...
-        throw Throwables.propagate(ex);
-      }
-      log.error(String.format("Could not get FileInfo for %s. Fetching FileStatus from remote file system :", backendPath.toString()), ex);
-      FileStatus fileStatus = parentFs.getFileStatus(backendPath);
+    if (CacheConfig.isFileStalenessCheckEnabled(conf)) {
+      FileStatus fileStatus = remoteFileSystem.getFileStatus(backendPath);
       this.fileSize = fileStatus.getLen();
       this.lastModified = fileStatus.getModificationTime();
     }
+    else {
```
stagraqubole, line 111: can remove duplication by changing this a bit:
```
fileSize = -1
if (!stalenessChecksEnabled):
    try:
        set fileSize with thrift call
    catch:
        log

if fileSize == -1:
    set fileSize and lastModified via getFileStatus
```

raunaqmorarka: Done

PR #384, opened 2020-05-15
Remove dependency on commons-lang

stagraqubole:
it might not be available in embedded mode e.g in case of Presto commons-land is shaded and its namespace is relocated.

Checked other dependencies that are marked `provided` scope. They are safe.

### DIFF
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/utils/DiskUtils.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/utils/DiskUtils.java
@@ -14,7 +14,6 @@ package com.qubole.rubix.bookkeeper.utils;
 
 import com.qubole.rubix.spi.CacheConfig;
 import com.qubole.rubix.spi.CacheUtil;
-import org.apache.commons.lang.math.NumberUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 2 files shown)


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/utils/DiskUtils.java
```diff
@@ -111,7 +110,14 @@ public static long getDirectorySizeInMB(File dirname)
...
       log.error(String.format("Exception while calculating the size of the folder %s with exception : %s", dirname.toString(), e.toString()));
     }
-    return NumberUtils.toLong(output.toString().split("\\s+")[0]) / 1024;
+    try {
+      return Long.parseLong(output.toString().split("\\s+")[0]) / 1024;
+    }
+    catch (NumberFormatException e)
+    {
+      log.warn(String.format("Unable to calculate size of directory %s, du output returned: %s", dirname, output.toString()));
+      return -1L;
```
harmandeeps, line 119: we should fix its caller `getCacheSizeMB` accordingly.

raunaqmorarka, line 119: In the library code the default value returned was 0.

stagraqubole: handling

stagraqubole: Yes, returning -1 to let the stats also have -1 (had missed the change in caller to forward -1) is better then returning 0 which implies dashboards on cache usage will show wrong data. With this we will know something is wrong in the system looking at stats/dashboards

PR #388, opened 2020-05-23
close remote input stream incase of exception in CachedRRC

rohangarg:


rohangarg:
Missed to fix this as well in the previous change.

PR #389, opened 2020-05-24
Cache config updates

raunaqmorarka:


rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -818,7 +823,11 @@ public static Configuration disableFSCaches(Configuration conf)
     conf.setBoolean("fs.s3n.impl.disable.cache", true);
     conf.setBoolean("fs.s3a.impl.disable.cache", true);
     conf.setBoolean("fs.wasb.impl.disable.cache", true);
+    conf.setBoolean("fs.wasbs.impl.disable.cache", true);
```
rohangarg, line 826: loop with a list of schemes would be cleaner.

raunaqmorarka: done

PR #391, opened 2020-05-26
Fixes to connection pool

stagraqubole:
1. Fix a connection leak in case of failed connection
2. Ensure a caller always gets a connection if the pool can grow

### DIFF
--- a/rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
+++ b/rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
@@ -21,6 +21,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.TimeUnit;
 
 import static com.google.common.base.Preconditions.checkState;
 
@@ -60,12 +61,55 @@ public class ObjectPoolPartition<T>
     }
   }
 
-  public BlockingQueue<Poolable<T>> getObjectQueue()
+  public void returnObject(Poolable<T> object)
   {
-    return objectQueue;
+    if (!objectFactory.validate(object.getObject())) {
+      log.debug(String.format("Invalid object for host %s removing %s ", object.getHost(), object));
+      decreaseObject(object);
+      // Compensate for the removed object. Needed to prevent endless wait when in parallel a borrowObject is called
+      increaseObjects(1);
+      return;
+    }
+
+    log.debug(String.format("Returning object %s to queue of host %s. Queue size: %d", object, object.getHost(), objectQueue.size()));
+    if (!objectQueue.offer(object)) {
+      log.warn("Created more objects than configured. Created=" + totalCount + " QueueSize=" + objectQueue.size());
+      decreaseObject(object);
+    }
   }
 
-  public synchronized int increaseObjects(int delta)
+  public Poolable<T> getObject(boolean blocking)
+  {
+    if (objectQueue.size() == 0) {
+      // increase objects and return one, it will return null if reach max size
+      int totalObjects = increaseObjects(this.config.getDelta());
+      if (totalObjects == 0) {
+        // Could not create objects, this is mostly due to connection timeouts hence no point blocking as there is not other producer of sockets
+        throw new RuntimeException("Could not add connections to pool");
+      }
+    }
+
+    Poolable<T> freeObject;
+    try {
+      if (blocking) {
+        freeObject = objectQueue.take();
+      }
+      else {
+        freeObject = objectQueue.poll(config.getMaxWaitMilliseconds(), TimeUnit.MILLISECONDS);
+        if (freeObject == null) {
+          throw new RuntimeException("Cannot get a free object from the pool");
+        }
+      }
+    }
+    catch (InterruptedException e) {
+      throw new RuntimeException(e); // will never happen
+    }
+
+    freeObject.setLastAccessTs(System.currentTimeMillis());
+    return freeObject;
+  }
+
+  private synchronized int increaseObjects(int delta)
   {
     int oldCount = totalCount;
     if (delta + totalCount > config.getMaxSize()) {
--- a/rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
+++ b/rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
@@ -22,6 +22,8 @@ import org.apache.commons.logging.LogFactory;
 
 import java.util.concurrent.BlockingQueue;
 
+import static com.google.common.base.Preconditions.checkState;
+
 /**
  * @author Daniel
  */
@@ -85,11 +87,21 @@ public class ObjectPoolPartition<T>
     return totalCount;
   }
 
-  public synchronized boolean decreaseObject(Poolable<T> obj)
+  public boolean decreaseObject(Poolable<T> obj)
   {
+    checkState(obj.getHost() != null, "Invalid object");
+    checkState(obj.getHost().equals(this.host),
+            "Call to free object of wrong partition, current partition=%s requested partition = %s",
+            this.host, obj.getHost());
+    objectRemoved();
     objectFactory.destroy(obj.getObject());
-    totalCount--;
+    obj.destroy();
     return true;
+  }
+
+  private synchronized void objectRemoved()
+  {
+    totalCount--;
   }
 
   public synchronized int getTotalCount()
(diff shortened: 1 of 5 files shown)


rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -58,40 +61,120 @@ public ObjectPoolPartition(ObjectPool<T> pool, PoolConfig config,
...
+      log.debug(String.format("Invalid object for host %s removing %s ", object.getHost(), object));
+      decreaseObject(object);
+      // Compensate for the removed object. Needed to prevent endless wait when in parallel a borrowObject is called
+      increaseObjects(1, false);
+      return;
+    }
+
+    log.debug(String.format("Returning object %s to queue of host %s. Queue size: %d", object, object.getHost(), objectQueue.size()));
+    try {
+      if (!objectQueue.offer(object, 5, TimeUnit.SECONDS)) {
```
harmandeeps, line 76: we can make the timeout configurable in PoolConfig.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -58,40 +61,120 @@ public ObjectPoolPartition(ObjectPool<T> pool, PoolConfig config,
...
+          if (objectToReturn == null && returnObject) {
+            objectToReturn = poolable;
+          }
+          else {
+            objectQueue.put(poolable);
+          }
           totalCount++;
         }
       }
       log.debug("Increased pool size by " + (totalCount - oldCount) + " to new size: " + totalCount + ", current queue size: " + objectQueue.size());
```
harmandeeps, line 147: should we add warn logs in case we are not able to increase the pool size?

harmandeeps: also, we can print delta as well, it may help us in debugging issues.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -58,40 +61,120 @@ public ObjectPoolPartition(ObjectPool<T> pool, PoolConfig config,
...
     }
-    catch (InterruptedException e) {
+    catch (Exception e) {
+      // objectToReturn is not on the queue hence untracked, clean it up before forwarding exception
```
harmandeeps, line 150: nit: we can add log.warn.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
stagraqubole: i avoided doing that, this is only during the return path and we have a fallback to destroy object if we ever hit this case. Additional config will never be tweaked by users.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
stagraqubole: adding

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
stagraqubole: adding

stagraqubole: done

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
stagraqubole: done

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -17,11 +17,14 @@
 package com.qubole.rubix.spi.fop;
 
 import com.qubole.rubix.spi.RetryingPooledBookkeeperClient;
+import com.sun.istack.Pool;
```
rohangarg: bad import?

stagraqubole: ya, is removed later but will remove from this commit

stagraqubole: done

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
rohangarg: This timeout should be < 5s I think - since it is not the same case as borrowing. I think there can be two cases where this can take time, either the number of objects created are > max size or the thread is unable to acquire the lock.
For the first case, we should instantly destroy the object and for the second one the timeout could be less.

stagraqubole: The second case shouldn't ever happen. I agree this could be very aggressive, as aggressive as try-and-return.  I will change this to `offer(E e)` call

stagraqubole: done

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -58,40 +61,123 @@ public ObjectPoolPartition(ObjectPool<T> pool, PoolConfig config,
...
+      log.debug(String.format("Invalid object for host %s removing %s ", object.getHost(), object));
+      decreaseObject(object);
+      // Compensate for the removed object. Needed to prevent endless wait when in parallel a borrowObject is called
+      increaseObjects(1, false);
+      return;
+    }
+
+    log.debug(String.format("Returning object %s to queue of host %s. Queue size: %d", object, object.getHost(), objectQueue.size()));
+    if (!objectQueue.offer(object)) {
+      log.error("Created more objects than configured: " + totalCount);
```
harmandeeps, line 76: nit: we can print queue size to for better debugging.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/Poolable.java
```diff
@@ -66,6 +66,13 @@ public void setLastAccessTs(long lastAccessTs)
     this.lastAccessTs = lastAccessTs;
   }
 
+  public void destroy()
```
harmandeeps, line 69: nit: we may call it 'reset' or something.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -58,40 +61,123 @@ public ObjectPoolPartition(ObjectPool<T> pool, PoolConfig config,
...
+            "Call to free object of wrong partition, current partition=%s requested partition = %s",
+            this.host, obj.getHost());
+    objectRemoved(obj);
     objectFactory.destroy(obj.getObject());
-    totalCount--;
+    obj.destroy();
     return true;
   }
 
+  private synchronized void objectRemoved(Poolable<T> obj)
```
harmandeeps, line 176: we can drop 'obj' now as it is not getting used. Also, may rename it to 'decreasseTotoalCount'.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/Poolable.java
harmandeeps: Also, as this. object won't be on queue anymore. Can't we get rid of this?

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
stagraqubole: Left that out as that is derivable from max-size config

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/Poolable.java
stagraqubole: This is here to catch any issues around reusing destroyed object in testing. destroy suits better as this object will not be reused again, reset implies this object is reusable.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
stagraqubole: Kept it as an extension point, if we want to do anything specific to obj in future. Will remove it for now and add when needed

stagraqubole: will retain the same name as we are informing PoolPartition via this method that an object is removed. It can do more than decreasing totalCount, just that right it is only doing that.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
stagraqubole: done

PR #392, opened 2020-05-26
Don't use shaded thrift imports

sopel39:
This allows to compile code in IntelliJ
and run "mvn test"


sopel39:
This doesn't solve the IntelliJ issue as mentioned in: https://stackoverflow.com/questions/30407065/getting-intellij-to-import-shaded-dependencies-in-a-multi-module-maven-project

PR #399, opened 2020-06-02
Use Presto NodeManager to obtain list of workers

sopel39:
Using NodeManager directly removes the need for
each Presto worker to talk directly to coordinator
via REST API (e.g when coordinator is secured)

Additionally, this change fixes bug when node list
on worker stops being updated after 300 seconds
once node determines it's a worker node.

### DIFF
--- a/rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
+++ b/rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
@@ -23,9 +23,13 @@ import com.google.gson.Gson;
 import com.qubole.rubix.common.utils.ClusterUtil;
 import com.qubole.rubix.spi.ClusterManager;
 import com.qubole.rubix.spi.ClusterType;
+import io.prestosql.spi.Node;
+import io.prestosql.spi.NodeManager;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
+
+import javax.annotation.Nullable;
 
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -46,6 +50,9 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static java.util.Objects.requireNonNull;
+
 /**
  * Created by stagra on 14/1/16.
  */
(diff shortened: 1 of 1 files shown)


rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
```diff
@@ -12,194 +12,41 @@
...
-  public static String serverPortConf = "caching.fs.presto-server-port";
+  public static void setNodeManager(NodeManager nodeManager)
+  {
+    PrestoClusterManager.nodeManager = requireNonNull(nodeManager, "nodeManager is null");
+  }
 
-  // Safe to use single instance of HttpClient since Supplier.get() provides synchronization
   @Override
-  public void initialize(Configuration conf)
+  public ClusterType getClusterType()
```
stagraqubole, line 40: We need to retain the existing code for standalone mode. Defining new cluster manager for embedded mode would require many changes, it would be simpler when we decouple FileSystems and ClusterManagers. For now, we need to have two suppliers of nodeList: native NodeManager based and /v1/node based, with native one, when set, getting preference over the other.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
```diff
@@ -209,13 +56,21 @@ public boolean isMaster()
...
+      return ImmutableList.of(nodeManager.getCurrentNode().getHost());
     }
-    catch (ExecutionException e) {
-      log.warn("Error fetching node list : ", e);
-    }
-    return null;
+
+    return workers.stream()
+            .map(Node::getHost)
+            .sorted()
```
stagraqubole, line 72: `getNodes` is called frequently and parallely too. Though the list to sort here will be small but would be better to keep a cache of node list here. We can refactor the code to have the cache logic common to both the providers.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
sopel39: > We need to retain the existing code for standalone mode.

Doesn't Rubix manage nodes themselves in standalone mode? Why `PrestoClusterManager` is needed then?

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
sopel39: > getNodes is called frequently and parallely too. Though the list to sort here will be small but would be better to keep a cache of node list here.

Makes sense.

sopel39:
applied comments @stagraqubole

sopel39:
tests passed locally

stagraqubole:
Merged, thanks.

PR #400, opened 2020-06-04
Provide cached locations in FileSystem#listLocatedStatus

sopel39:
Making CachingFileSystem a FilterFileSystem caused listLocatedStatus
to be delegated to DistributedFileSystem directly (instead of using
default FileSystem#listLocatedStatus) implementation. This caused
produced file statuses to contain remote BlockLocation (instead of
cached ones).

This change overrides listLocatedStatus method so that it returns
cached BlockLocations. Additionally, it does it in iterative fashion
instead of fetching all locations eagerly (as default FileSystem#listLocatedStatus)
implementation does.

### DIFF
--- a/rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
+++ b/rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
@@ -29,7 +29,9 @@ import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.FilterFileSystem;
+import org.apache.hadoop.fs.LocatedFileStatus;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.RemoteIterator;
 import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.util.Progressable;
 import org.weakref.jmx.MBeanExporter;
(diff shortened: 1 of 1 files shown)


sopel39:
@stagraqubole tests passed locally

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -319,6 +321,33 @@ private URI getOriginalURI(URI actualURI)
     return URI.create(getScheme() + "://" + actualURI.getAuthority());
   }
 
+  @Override
+  public RemoteIterator<LocatedFileStatus> listLocatedStatus(Path path)
+          throws IOException
+  {
+      return new RemoteIterator<LocatedFileStatus>()
+      {
+          private final RemoteIterator<LocatedFileStatus> stats = fs.listLocatedStatus(path);
```
rohangarg, line 330: also, any reasons for not doing `listStatus`? we're anyways not using the block locations returned.

sopel39: > any reasons for not doing listStatus? we're anyways not using the block locations returned.

List status fetches all statuses at once. This could be problematic if directory has thousands of files (or you don't need to read all files). E.g Presto uses `listLocatedStatus` in an iterative way to to provide splits.

rohangarg: ok, that's makes sense. 👍

PR #401, opened 2020-06-05
Reduce logging level for FileNotFoundExceptions

sopel39:
In parallel warmup mode file might be gone by the
time Rubix tries to download it. Since this is something
that is expected, it doesn't need to be logged as error.


sopel39:
FYI: @stagraqubole

PR #407, opened 2020-06-12
Fix over-estimation of Cache size

### LINKED ISSUE #404: Over-estimation of cache size for small files
Because we increment the size of the cache entry in the increments of block size, if cache has lot of small files then we can be over estimating the disk space used. For the default 1MB block size, for a file of size 0.1MB in cache we account 1MB as occupied space.
We know the fileSize at the time we are updating the occupied size in `setAllCached`, we can use that info to cap the value accounted.

stagraqubole:
fixes #404

Last Block of files is usually not aligned to Block boundary, 1MB. But
we were adding the full weight of Block, 1MB, to the Cache for such blocks.
This lead to over-estimation of Cache size and hence early evictions.

This change limits the max size of file computed by the actual file size.
Also, with this change, the weight of each file in Cache is represent by it's size in KB
instead of MB. Which means we cannot support file > 1TB in size, which is a big
enough.

### DIFF
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
@@ -59,7 +59,6 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
-import java.nio.ByteBuffer;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -73,6 +72,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import static com.google.common.base.Preconditions.checkState;
+import static com.qubole.rubix.spi.utils.DataSizeUnits.BYTES;
 import static com.qubole.rubix.common.metrics.BookKeeperMetrics.CacheMetric.CACHE_AVAILABLE_SIZE_GAUGE;
 import static com.qubole.rubix.common.metrics.BookKeeperMetrics.CacheMetric.CACHE_EVICTION_COUNT;
 import static com.qubole.rubix.common.metrics.BookKeeperMetrics.CacheMetric.CACHE_EXPIRY_COUNT;
@@ -626,7 +626,7 @@ public abstract class BookKeeper implements BookKeeperService.Iface
     for (int d = 0; d < cacheDiskCount; d++) {
       avail += new File(CacheUtil.getDirPath(d, conf)).getUsableSpace();
     }
-    avail = DiskUtils.bytesToMB(avail);
+    avail = BYTES.toMB(avail);
     log.info("total free space " + avail + "MB");
 
     // In corner cases evictions might not make enough space for new entries
--- a/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
+++ b/rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
@@ -72,7 +72,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import static com.google.common.base.Preconditions.checkState;
-import static com.qubole.rubix.spi.utils.DataSizeUnits.BYTES;
 import static com.qubole.rubix.common.metrics.BookKeeperMetrics.CacheMetric.CACHE_AVAILABLE_SIZE_GAUGE;
 import static com.qubole.rubix.common.metrics.BookKeeperMetrics.CacheMetric.CACHE_EVICTION_COUNT;
 import static com.qubole.rubix.common.metrics.BookKeeperMetrics.CacheMetric.CACHE_EXPIRY_COUNT;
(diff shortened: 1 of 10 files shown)


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -226,7 +227,7 @@ public Integer getValue()
       @Override
       public Long getValue()
```
harmandeeps, line 228: should we add 'KB' in metric as well?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -247,9 +248,9 @@ void deleteFiles(Cache<String, FileMetadata> cache)
     }
   }
 
-  public int getWeight(Configuration conf)
+  // Returns the current downloaded fileSize in KB
+  public int getWeight()
```
harmandeeps, line 252: same, should we add KB here as well?

harmandeeps: Also, should we get rid of 'getCurrentFileSize()'?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
raunaqmorarka: CACHE_AVAILABLE_SIZE_GAUGE needs to be updated from `rubix.bookkeeper.gauge.available_cache_size_mb` to `rubix.bookkeeper.gauge.available_cache_size_kb`

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -445,6 +446,10 @@ public void setAllCached(String remotePath, long fileLength, long lastModified,
       OptionalInt updatedBlocks = md.setBlocksCached(startBlock, endBlock);
       if (updatedBlocks.isPresent()) {
         long currentFileSize = md.incrementCurrentFileSize(updatedBlocks.getAsInt() * CacheConfig.getBlockSize(conf));
+        // CurrentFileSize as per Blocks' based computation can cross actual file size
+        // This can only happen when the last block of file is not completely full
+        // as it doesnt align to the block boundary
+        currentFileSize = currentFileSize > fileLength ? fileLength : currentFileSize;
```
raunaqmorarka, line 452: Math.min(currentFileSize, fileLength)

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -453,6 +458,12 @@ public void setAllCached(String remotePath, long fileLength, long lastModified,
     }
   }
 
+  @VisibleForTesting
+  public long getTotalCacheWeight()
+  {
+    return fileMetadataCache.asMap().values().stream().map(md -> (long) md.getWeight()).reduce(0L, Long::sum);
```
raunaqmorarka, line 464: `fileMetadataCache.asMap().values().stream().mapToLong(FileMetadata::getWeight).sum();`

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
raunaqmorarka, line 252: int -> long

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
```diff
@@ -247,9 +248,9 @@ void deleteFiles(Cache<String, FileMetadata> cache)
...
   {
-    // this will return the current downloaded size of the file as weight.
-    return (int) (currentFileSize / 1024 / 1024);
+    return (int) BYTES.toKB(currentFileSize);
```
raunaqmorarka, line 254: Do we even need to change this ?
Can use Math.toIntExact instead of cast

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/utils/DiskUtils.java
```diff
@@ -130,6 +109,6 @@ public static int getUsedSpaceMB(Configuration conf)
       File localPath = new File(CacheUtil.getDirPath(d, conf));
       used += localPath.getTotalSpace() - localPath.getUsableSpace();
     }
-    return (int) bytesToMB(used);
+    return (int) BYTES.toMB(used);
```
raunaqmorarka, line 112: getUsedSpaceMB seems to be unused, can remove if not needed

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
stagraqubole: I think better to revert this and keep user visible things in MB. Easier to look at values in MB than in KB, will change this part.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
stagraqubole: changing

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/FileMetadata.java
stagraqubole: storing weigh in KB is internal detail and not needed to be added to method name.

getCurrentFileSize is being used by tests, are you suggesting to change those tests?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -622,17 +633,17 @@ public FileInfo load(String key) throws Exception
     for (int d = 0; d < CacheUtil.getCacheDiskCount(conf); d++) {
       avail += new File(CacheUtil.getDirPath(d, conf)).getUsableSpace();
     }
-    avail = DiskUtils.bytesToMB(avail);
+    avail = BYTES.toMB(avail);
     log.info("total free space " + avail + "MB");
```
raunaqmorarka, line 637: Can move this log to line 647 and also log `totalAvailableForCacheInMB` there

PR #413, opened 2020-06-17
Do not obtain localPath when caching data on master is disabled

sopel39:
This should prevent errors like https://github.com/prestosql/presto/pull/3987#issuecomment-642225940 when caching data on master is disabled. Cache itself needs to be enabled on master because coordinator needs to know where to schedule splits.


stagraqubole:
This variable is getting dropped in https://github.com/qubole/rubix/pull/381/files but I am merging this for just in case #381 is delayed

PR #415, opened 2020-06-22
Don't shade GCS connector

sopel39:


stagraqubole:
@sopel39 were you able to get queries working with this change? In rubix-presto module we exclude `<exclude>com.google.cloud.**</exclude>` for GCS and I remember that worked well and iirc just doing `<exclude>com.google.cloud</exclude>` did not help

sopel39:
@stagraqubole let me check. I will get back to you

sopel39:
@stagraqubole you were right. I've updated PR and now `io.prestosql.plugin.hive.rubix.TestRubixCaching#testFileSystemBindings` passes

PR #423, opened 2020-07-07
Use bounded thread pools for LocalDataTransferServer and BookKeeperServer

raunaqmorarka:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -174,7 +179,11 @@ public LocalServer(Configuration conf, BookKeeperFactory bookKeeperFactory)
     @Override
     public void run()
     {
-      ExecutorService threadPool = Executors.newCachedThreadPool();
+      ExecutorService threadPool = new ThreadPoolExecutor(
+              0, getLocalTransferServerMaxThreads(conf),
+              60L, TimeUnit.SECONDS,
```
harmandeeps, line 184: nit: we can create config for `keepAliveTime`.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -157,7 +158,7 @@
   private static final int DEFAULT_REMOTE_FETCH_PROCESS_INTERVAL = 10000; // ms
   private static final int DEFAULT_REMOTE_FETCH_THREADS = 10;
   private static final boolean DEFAULT_RUBIX_ON_MASTER = false;
-  private static final int DEFAULT_SERVER_MAX_THREADS = Integer.MAX_VALUE;
+  private static final int DEFAULT_SERVER_MAX_THREADS = Runtime.getRuntime().availableProcessors() * 16;
```
stagraqubole, line 161: This is too low. With connection pooling, each connection has its own thread and in case of idle periods it won't consume resources. Lowering it to such a low value is going to disable connection pooling in a way. We should keep it higher, imo few thousands.

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
```diff
@@ -174,7 +179,11 @@ public LocalServer(Configuration conf, BookKeeperFactory bookKeeperFactory)
...
+              new SynchronousQueue<>(),
+              threadsNamed("lds-pool-%s"));
```
harmandeeps, line 186: We use `lds-pool` keyword for connection pool. I guess we may rename it to something else to remove any confusion.

harmandeeps: nevermind, looks good. thought something else.

raunaqmorarka: will rename to `lds-worker-%s`

stagraqubole:
Changing BookKeeperServer separately?

rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/LocalDataTransferServer.java
raunaqmorarka: Thread creation is fast and we're going through an idle timeout on client side anyway before the thread on server is released. So keeping this at 1m for simplicity.

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
raunaqmorarka: Increased to 4k

PR #424, opened 2020-07-10
Avoid unnecessary lookup in fileMetadataCache before invalidation

raunaqmorarka:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -750,62 +744,22 @@ public FileInfo load(String s) throws Exception
...
-            case SIZE:
-              cacheEvictionCount.inc();
-              break;
-            case EXPIRED:
-              cacheExpiryCount.inc();
-              break;
-            default:
-              break;
-          }
+      md.closeAndCleanup(notification.getCause(), fileMetadataCache);
```
stagraqubole, line 747: Can you add the logs in `FileMetada.closeAndCleanup` when `mdFile.delete()` or `localFile.delete()` return false

raunaqmorarka: Done

PR #426, opened 2020-07-13
Use semaphore in ObjectPool

harmandeeps:
Draft changes to use semaphore in ObjectPool and some minor fixes.


rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -78,85 +82,55 @@ public void returnObject(Poolable<T> object)
...
-      if (totalCount == 0) {
-        // Could not create objects, this is mostly due to connection timeouts hence no point blocking as there is not other producer of sockets
-        throw new RuntimeException("Could not add connections to pool");
-      }
-      // else wait for a connection to get free
+    try {
+      takeSemaphore.tryAcquire(config.getMaxWaitMilliseconds(), TimeUnit.MILLISECONDS);
+    }
+    catch (InterruptedException e) {
+      e.printStackTrace();
```
stagraqubole, line 92: remove e.printStackTrace

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -78,85 +82,55 @@ public void returnObject(Poolable<T> object)
...
+      throw new RuntimeException("Cannot get a free object from the pool");
     }
 
-    Poolable<T> freeObject;
-    try {
-      if (blocking) {
-        freeObject = objectQueue.take();
+    Poolable<T> object;
+    if (blocking) {
```
stagraqubole, line 97: crossing the semaphore acquisition => thread should either get an existing connection from pool or create a new one. `blocking` doesnt apply at this point.
Instead, if `blocking` is set then you need to call `takeSemaphore.acquire` and in the other case call `takeSemaphore.tryAcquire`

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -78,85 +82,55 @@ public void returnObject(Poolable<T> object)
...
     }
-
-    freeObject.setLastAccessTs(System.currentTimeMillis());
-    return freeObject;
+    object.setLastAccessTs(System.currentTimeMillis());
+    return object;
   }
 
-  private Poolable<T> increaseObjects(int delta, boolean returnObject)
+  private Poolable<T> tryGetObject(Poolable<T> poolable)
```
stagraqubole, line 113: Dont take `Poolable` as arg, instead do `objectQueue.poll()` inside this method

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -78,85 +82,55 @@ public void returnObject(Poolable<T> object)
...
         }
-      }
-
-      if (delta > 0 && (totalCount - oldCount) == 0) {
-        log.warn(String.format("Could not increase pool size. Pool state: totalCount=%d queueSize=%d delta=%d", totalCount, objectQueue.size(), delta));
-      }
-      else {
         log.debug(String.format("Increased pool size by %d, to new size: %d, current queue size: %d, delta: %d",
-                totalCount - oldCount, totalCount, objectQueue.size(), delta));
+                1, totalCount, objectQueue.size(), 1));
```
stagraqubole, line 122: Remove constant `1`, change message to `Added a connection, new size: ..`

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -78,85 +82,55 @@ public void returnObject(Poolable<T> object)
...
       }
-    }
-    catch (Exception e) {
-      log.warn(String.format("Unable to increase pool size. Pool state: totalCount=%d queueSize=%d delta=%d", totalCount, objectQueue.size(), delta), e);
-      // objectToReturn is not on the queue hence untracked, clean it up before forwarding exception
-      if (objectToReturn != null) {
-        objectFactory.destroy(objectToReturn.getObject());
-        objectToReturn.destroy();
+      catch (Exception e) {
+        log.warn(String.format("Unable to increase pool size. Pool state: totalCount=%d queueSize=%d delta=%d", totalCount, objectQueue.size(), 1), e);
```
stagraqubole, line 125: same here, remove `1`

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -78,85 +82,55 @@ public void returnObject(Poolable<T> object)
...
+      throw new RuntimeException("Cannot get a free object from the pool");
     }
 
-    Poolable<T> freeObject;
-    try {
-      if (blocking) {
-        freeObject = objectQueue.take();
+    Poolable<T> object;
+    if (blocking) {
+      try {
```
stagraqubole, line 98: move code below into try-catch and release semaphore in case of exception

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -63,6 +66,7 @@ public ObjectPoolPartition(ObjectPool<T> pool, PoolConfig config,
 
   public void returnObject(Poolable<T> object)
   {
+    takeSemaphore.release();
```
stagraqubole, line 72: remove `takeSemaphore.release()` here, move rest of the code in try-finally and call `takeSemaphore.release()` in the finally block. Otherwise current code opens up the case of going beyond max connections

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -56,11 +56,14 @@ public ObjectPoolPartition(ObjectPool<T> pool, PoolConfig config,
...
     this.takeSemaphore = new Semaphore(config.getMaxSize(), true);
     for (int i = 0; i < config.getMinSize(); i++) {
-      T object = objectFactory.create(host, socketTimeout, connectTimeout);
-      if (object != null) {
+      try {
+        T object = objectFactory.create(host, socketTimeout, connectTimeout);
         objectQueue.add(new Poolable<>(object, pool, host));
         totalCount++;
       }
+      catch (Exception e) {
```
stagraqubole, line 64: move full `for` loop inside try-catch. Otherwise we will generate one warn log for each trial

stagraqubole: this is log is actually not needed, factories are logging cases of exception

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: yeah, this are draft changes, I will remove all these contents and redundant code in the next iteration.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: fixed.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: good point, fixed.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: fixed.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: fixed.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: fixed.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: fixed.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: makes sense, fixed.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: removed

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -50,115 +53,83 @@ public ObjectPoolPartition(ObjectPool<T> pool, PoolConfig config,
...
-        }
+        takeSemaphore.tryAcquire(config.getMaxWaitMilliseconds(), TimeUnit.MILLISECONDS);
+        object = tryGetObject();
       }
+      object.setLastAccessTs(System.currentTimeMillis());
     }
-    catch (InterruptedException e) {
-      throw new RuntimeException(e); // will never happen
+    catch (Exception e) {
+      takeSemaphore.release();
```
rohangarg, line 106: is it ok to release the semaphore even if the thread was interrupted while trying to acquire the lock from semaphore?

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -50,115 +53,83 @@ public ObjectPoolPartition(ObjectPool<T> pool, PoolConfig config,
...
+        return;
+      }
 
-    log.debug(String.format("Returning object: %s to queue. Queue size: %d", object, objectQueue.size()));
-    if (!objectQueue.offer(object)) {
-      log.warn("Created more objects than configured. Created=" + totalCount + " QueueSize=" + objectQueue.size());
-      decreaseObject(object);
+      log.debug(String.format("Returning object: %s to queue. Queue size: %d", object, objectQueue.size()));
+      if (!objectQueue.offer(object)) {
+        log.warn("Created more objects than configured. Created=" + totalCount + " QueueSize=" + objectQueue.size());
```
rohangarg, line 82: is it ok to release the semaphore incase more objects were created than `maxSize`

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -50,115 +53,83 @@ public ObjectPoolPartition(ObjectPool<T> pool, PoolConfig config,
...
-      // else wait for a connection to get free
-    }
-
-    Poolable<T> freeObject;
+    Poolable<T> object;
     try {
       if (blocking) {
-        freeObject = objectQueue.take();
+        takeSemaphore.acquire();
+        object = objectQueue.take();
```
rohangarg, line 97: what's the reason for not calling `tryGetObject` here?

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -35,10 +37,11 @@
   private final PoolConfig config;
   private final BlockingQueue<Poolable<T>> objectQueue;
   private final ObjectFactory<T> objectFactory;
-  private int totalCount;
   private final String host;
   private final int socketTimeout;
   private final int connectTimeout;
+  private final Semaphore takeSemaphore;
+  private final AtomicInteger totalCount;
```
rohangarg, line 44: is it possible to remove `totalCount` and rather use `takeSemaphore#permits`

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -174,20 +145,20 @@ public boolean decreaseObject(Poolable<T> obj)
...
 
-  private synchronized void objectRemoved()
+  private void objectRemoved()
   {
-    totalCount--;
+    totalCount.decrementAndGet();
   }
 
-  public synchronized int getTotalCount()
+  public int getTotalCount()
```
rohangarg, line 153: should this be `getActiveConnectionCount` (the var too)?

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -50,115 +53,83 @@ public ObjectPoolPartition(ObjectPool<T> pool, PoolConfig config,
...
       }
       else {
-        freeObject = objectQueue.poll(config.getMaxWaitMilliseconds(), TimeUnit.MILLISECONDS);
-        if (freeObject == null) {
-          throw new RuntimeException("Cannot get a free object from the pool");
-        }
+        takeSemaphore.tryAcquire(config.getMaxWaitMilliseconds(), TimeUnit.MILLISECONDS);
+        object = tryGetObject();
       }
+      object.setLastAccessTs(System.currentTimeMillis());
```
rohangarg, line 103: why is this set while giving the object? what if this object is being used for more than the max-idle time?
shouldn't this be set to some sentinel value while giving the object, and to a timestamp while release of the object of the pool?

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
stagraqubole: It is not. We should call release only when we acquire a permit

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
stagraqubole: We should. Or we should remove the code path for `blocking=true` case as it is not applicable for us and it is only complicating maintenance.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
stagraqubole: We should fix this separately, I will open a issue. Right now it works because `thrift call time` is much less than `idle timeout`.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
stagraqubole: No, this needs a change

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: tryGetObject call object.Queue.poll() which is non blocking. But, in case of blocking mode, though there is semaphore acquire blocking call, I just a called 'take' to be on safe side.

I guess better removing this blocking mode as we are not using it anymore in the code.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: What do you mean by max-idle time? If you are referring scavenging the under connection, then in this case , object won't be present in the queue to get scavenged. 

All the objects are inherently same (don't have any state properties associated), setting timestamp while getting or returning shouldn't matter much. Or am I missing something?

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: yeah, this needs to be fixed. will fix it in the next iteration.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: maybe I am missing something. Theoretically, this case should happen as we are controlling parallelism via counter-based semaphore. 
suppose, this happens, then why shouldn't we release the semaphore as the returning thread is done with the object? won't it give a chance to other waiting threads?

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: will change this.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: yeah, will change in next iteration.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
rohangarg: if we are controlling parallelism strictly via counter-based semaphore, then this is an error state. We should error out if this happens instead.
I don't think it makes sense to operate a semaphore if the number of objects can be larger than permits (it can lead to wrong calculations).

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
rohangarg: Agree with Shubham as to why it works for now.

> setting timestamp while getting or returning shouldn't matter much.

Ok, if this will not be a part of scavenger thread, then doesn't make sense to set the last access time in `release` method? `borrow` is logically not the last access time according to me.

rohangarg:
Also, can you please the exact motivation for using semaphore in the object pool?
Were things like atomic counter + mutex locking or blocking queue not enough for controlling concurrency? (and why if they were not).

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: yeah, got it. makes sense. for example, if in a query burst all objects are taken for 20 minutes (suppose) and the released. Scanvenger will think these objects are not getting used and scavenge. 

We can take this in a separate commit as subhuam mentioned.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: I checked it is not possible. We initialize the pool with min size without acquiring semaphore.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: fixed.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: yes, true. throwing error in such a case.

harmandeeps:
> Also, can you please the exact motivation for using semaphore in the object pool?
> Were things like atomic counter + mutex locking or blocking queue not enough for controlling concurrency? (and why if they were not).

changed the description. please check.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
```diff
@@ -167,27 +96,69 @@ public boolean decreaseObject(Poolable<T> obj)
...
     obj.destroy();
     return true;
   }
 
-  private synchronized void objectRemoved()
+  public Poolable<T> getObject()
+  {
+    Poolable<T> object;
+    try {
+      takeSemaphore.tryAcquire(config.getMaxWaitMilliseconds(), TimeUnit.MILLISECONDS);
```
stagraqubole, line 110: Return value denotes timeout. If we timeout then we need to return instead of going ahead without permit, that can cause creating more than max connections for some time. At the time of returning we would destroy extra ones and then double decrement semaphore.
Please add UTs around this to capture different cases.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPool.java
```diff
@@ -78,8 +76,11 @@ public void registerHost(String host)
     }
     log.debug(this.name + " : Borrowing object for partition: " + host);
     for (int i = 0; i < 3; i++) { // try at most three times
```
stagraqubole, line 78: Remove this retry, better exit fast and fallback to direct reads.

stagraqubole (changes requested):

rubix-spi/src/main/java/com/qubole/rubix/spi/CacheConfig.java
```diff
@@ -310,7 +308,7 @@ public static int getTransportPoolMaxWait(Configuration conf)
 
   public static int getScavengeInterval(Configuration conf)
   {
-    return conf.getInt(KEY_POOL_MAX_WAIT_TIMEOUT, DEFAULT_SCAVENGE_INTERVAL);
+    return conf.getInt(KEY_POOL_SCAVENGER_INTERVAL, DEFAULT_SCAVENGE_INTERVAL);
```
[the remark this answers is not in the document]
harmandeeps: created a separate commit.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPool.java
harmandeeps: removed.

rubix-spi/src/main/java/com/qubole/rubix/spi/fop/ObjectPoolPartition.java
harmandeeps: Added logic to check the return type, if it is false (lock not acquired), return null. Also, added UT around the semaphore logic. Please have a look.

PR #434, opened 2020-07-24
Initialize stats in setLocalBookKeeper

losipiuk:


losipiuk:
Not tested yet.

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -99,26 +99,36 @@ private static void initialize(Configuration conf, ClusterType clusterType)
...
+      if (statsInitialized.get()) {
+        return;
       }
-      if (!ManagementFactory.getPlatformMBeanServer().isRegistered(new ObjectName(detailedStatsMBeanName))) {
-        exporter.export(detailedStatsMBeanName, new DetailedCachingFileSystemStats(stats, bookKeeperFactory, conf));
+
+      MBeanExporter exporter = new MBeanExporter(ManagementFactory.getPlatformMBeanServer());
+      stats = new CachingFileSystemStatsProvider();
+      try {
+        if (ManagementFactory.getPlatformMBeanServer().isRegistered(new ObjectName(statsMBeanName))) {
```
sopel39, line 114: if the initialization is idempotent, it doesn't make sense to unregister stats

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -172,14 +182,15 @@ public FileSystem getRemoteFileSystem()
     return fs;
   }
 
-  public static void setLocalBookKeeper(BookKeeperService.Iface bookKeeper, String statsMbeanSuffix)
+  public static void setLocalBookKeeper(BookKeeperService.Iface bookKeeper, String statsMbeanSuffix, Configuration configuration)
   {
+    checkState(!initialized.get(), "LocalBookKeeper should be set before opening up the Filesystem to clients");
```
sopel39, line 187: should we just remove this check and reinitialize stats with new bookkeeper?

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -85,7 +86,6 @@ private static void initialize(Configuration conf, ClusterType clusterType)
       if (initialized.get()) {
         return;
       }
-
       initializeStats(conf);
```
sopel39, line 89: I don't think stats should be initialized on caching FS initialization. In Presto we create more than one `CachingFileSystem` (with different configs: one with cache enabled, one with cache disabled, one for bookeeper/filedownloader). Since stats are dependent on configuration, we are not really sure which config would be used for metrics.

It seems that `setLocalBookKeeper` should be the method initializing/reinitializing stats.

stagraqubole: We setLocalBookKeeper version will always create the stats first in embedded mode as we do not open up presto for traffic before BookKeeper is initialized. For safety, we can remove the check later in setLocalBookKeeper and ensure we always override stats through it.

The part about initializing stats via FS need to stay because in standalone mode that is the only first point where `Configuration` object is accessible.

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
stagraqubole: Though this check is safe with the way Rubix is used in Presto right now: `setLocalBookKeeper` is called before FS initialize; But for safety we can remove this check and always override stats from here.

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
stagraqubole: It does with current changes but given that we are removing the `checkState(!initialized.get()` from `setLocalBookKeeper` => we are allowing re-init of stats so we will need to keep this.

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
sopel39: > The part about initializing stats via FS need to stay because in standalone mode that is the only first point where Configuration object is accessible.

Maybe stats could be initialized in CachingFS only in non-embedded mode? This could be checked from passed `Configuration`

sopel39:
After discussion with @stagraqubole we decided to:
1. make `setLocalBookKeeper` (re)initialize stats
2. disable automatic stats initialization in CachingFS in embedded mode

losipiuk:
@stagraqubole I updated the PR per what you discussed with @sopel39.

I also added `CachingFileSystem.deinitialize` method to be used in tests only to address the problem with running tests with multiple configurations, which was blocket by introduction of intialization guarding.

Please let me know if that if fine for you and can we get rubix 0.3.16 with those changes out.

rubix-core/src/main/java/com/qubole/rubix/core/CachingFileSystem.java
```diff
@@ -97,6 +102,16 @@ private static void initialize(Configuration conf, ClusterType clusterType)
     }
   }
 
+  // Deinitialization rouitne for unit test which needs reinitialize CachingFileSystem multiple times with
+  // different configurations.
+  @VisibleForTesting
+  public static void deinitialize() {
+    clusterManager = null;
+    statsMBeanName = "rubix:name=stats";
```
stagraqubole, line 110: nit: these names can be moved to constants

losipiuk:
@stagraqubole thanks for review. Comment addressed.
If possible please merge and trigger release of Rubix. Then I will send out PR which updates version of Rubix in https://github.com/prestosql/presto.

stagraqubole:
@losipiuk released 0.3.16

losipiuk:
> @losipiuk released 0.3.16

Great. THanks a lot!

PR #437, opened 2020-08-12
ClusterManager refactoring and adding ability to provide custom hostname by ClusterManager implementations

### LINKED ISSUE #436: Extend ClusterManager to return current node or current node index
Currently, BookKeeper uses hostname or default IP to calculate index of current node vs all Rubix nodes. However, this doesn't always work in more sophisticated deployments, where Presto nodes use neither to announce themselves to coordinator. Therefore `ClusterManager` should provide `getCurrentNode` or `getCurrentNodeIndex` that would work in such cases.

stagraqubole:
Main changes:
1. Offload all cluster related management like node-list, currentNodeIndex from BookKeeper to ClusterManager class
2. Simplifies implementation of ClusterManager by moving node-list caching into common place in ClusterManager
3. Adds logic to let ClusterManager implementations provide custom hostname/hostAddress to find currentNodeIndex. This fixes #436 as PrestoClusterManager will now return hostname and hostAddress that is registered with NodeManager


rubix-presto/src/main/java/com/qubole/rubix/presto/PrestoClusterManager.java
```diff
@@ -69,167 +58,108 @@ public void initialize(Configuration conf)
...
+      URL allNodesRequest = getNodeUrl();
+      URL failedNodesRequest = getFailedNodeUrl();
+
+      HttpURLConnection allHttpCon = (HttpURLConnection) allNodesRequest.openConnection();
+      allHttpCon.setConnectTimeout(500); //ms
+      allHttpCon.setRequestMethod("GET");
+
+      int allNodesResponseCode = allHttpCon.getResponseCode();
+
+      StringBuffer allResponse = new StringBuffer();
```
raunaqmorarka, line 76: `StringBuilder` instead of `StringBuffer`

rubix-spi/src/main/java/com/qubole/rubix/spi/ClusterManager.java
```diff
@@ -31,53 +49,143 @@
...
+    return nodeHostAddress;
+  }
+
+  private List<String> getNodesAndUpdateState()
+  {
+    List<String> nodes = getNodesInternal();
+    if (nodes == null || nodes.size() == 0) {
+      log.error("Could not initialize as no cluster node is found");
+    }
+    else if (nodes.indexOf(getCurrentNodeHostname()) >= 0) {
```
raunaqmorarka, line 84: We could avoid doing the `indexOf` lookup twice here

rubix-hadoop2/src/main/java/com/qubole/rubix/hadoop2/Hadoop2ClusterManager.java
```diff
@@ -51,89 +39,41 @@ public void initialize(Configuration conf)
...
 
   @Override
-  public List<String> getNodes()
+  public List<String> getNodesInternal()
   {
     try {
-      return nodesCache.get("nodeList");
+      List<Hadoop2ClusterManagerUtil.Node> allNodes = Hadoop2ClusterManagerUtil.getAllNodes(yconf);
+      if (allNodes == null) {
+        return null;
```
harmandeeps, line 50: we can return an empty list so that the caller doesn't need to do the null check.

rubix-hadoop2/src/main/java/com/qubole/rubix/hadoop2/Hadoop2ClusterManager.java
```diff
@@ -51,89 +39,41 @@ public void initialize(Configuration conf)
...
+      if (allNodes.isEmpty()) {
+        return ImmutableList.of();
+      }
+
+      Set<String> hosts = new HashSet<>();
+      for (Hadoop2ClusterManagerUtil.Node node : allNodes) {
+        String state = node.getState();
+        log.debug("Hostname: " + node.getNodeHostName() + "State: " + state);
+        //keep only healthy data nodes
+        if (state.equalsIgnoreCase("Running") || state.equalsIgnoreCase("New") || state.equalsIgnoreCase("Rebooted")) {
```
harmandeeps, line 62: we may use enums for states.

rubix-hadoop2/src/main/java/com/qubole/rubix/hadoop2/Hadoop2ClusterManager.java
harmandeeps: nevermind just saw the comment related to the distinction of the empty and null cases.

rubix-spi/src/main/java/com/qubole/rubix/spi/ClusterManager.java
```diff
@@ -46,6 +49,11 @@
  */
 public abstract class ClusterManager
 {
+  private static Log log = LogFactory.getLog(ClusterManager.class);
+
+  private int currentNodeIndex = -1;
+  private String nodeHostname;
```
harmandeeps, line 55: should we initialize to empty string to avoid any null failures?

rubix-spi/src/main/java/com/qubole/rubix/spi/ClusterManager.java
```diff
@@ -160,10 +160,32 @@ public int getNodeIndex(int numNodes, String key)
...
+  }
+
+  public static class ClusterInfo
+  {
+    private final List<String> nodes;
+    private final int currentNodeIndex;
+
+    public ClusterInfo (List<String> nodes, int currentNodeIndex)
+    {
+      this.nodes = nodes;
```
harmandeeps, line 177: should we add checks here?

harmandeeps: nit: space between ClusterInfo and (

rubix-spi/src/main/java/com/qubole/rubix/spi/ClusterManager.java
```diff
@@ -160,10 +160,32 @@ public int getNodeIndex(int numNodes, String key)
...
+      this.currentNodeIndex = currentNodeIndex;
+    }
+
+    public List<String> getNodes ()
```
harmandeeps, line 181: same space issue.

rubix-spi/src/main/java/com/qubole/rubix/spi/ClusterManager.java
```diff
@@ -160,10 +160,32 @@ public int getNodeIndex(int numNodes, String key)
...
+    {
+      return nodes;
+    }
+
+    public int getCurrentNodeIndex ()
```
harmandeeps, line 186: same.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
```diff
@@ -182,6 +183,32 @@ public void initialize(Configuration conf)
     return workers;
   }
 
+  @Override
+  protected String getCurrentNodeHostname()
+  {
+    if (nodeManager != null) {
+      nodeManager.getCurrentNode().getHost();
```
harmandeeps, line 190: missing return statement?

rubix-hadoop2/src/main/java/com/qubole/rubix/hadoop2/Hadoop2ClusterManager.java
stagraqubole: These are not Rubix states but coming from Hadoop. There are enums in Hadoop for NodeState but during serialisation they are converted to string. We can add logic to convert strings back into enums but that should be outside this change (this is existing code).

rubix-spi/src/main/java/com/qubole/rubix/spi/ClusterManager.java
stagraqubole: Actually we should throw exception in hostname/address initialisation itself otherwise we will hit NPE on nodeCache later. Will add a commit for that. Will add requireNonNull on accessors of these variables

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
stagraqubole: yes

rubix-presto/src/main/java/com/qubole/rubix/presto/PrestoClusterManager.java
```diff
@@ -65,171 +55,113 @@
...
+          }
+          catch (IOException e) {
+            throw new IOException(e);
+          }
+          finally {
+            in.close();
+          }
+        }
+        else {
+          log.warn(String.format("v1/node failed with code"));
```
suryanshagnihotri, line 97: code = allNodesResponseCode?

stagraqubole: will change, I avoided changing any existing code.

PR #438, opened 2020-08-18
Add stats about the total data read from source (client + BookKeeper)

stagraqubole:
c.c. @sopel39


rubix-core/src/main/java/com/qubole/rubix/core/DetailedCachingFileSystemStats.java
```diff
@@ -130,4 +130,12 @@ public int getcorrupted_file_count()
     {
         return statsProvider.getStats().getCorruptedFileCount();
     }
+
+    @Managed(description = "Total data read from source by the system")
+    // getMB_read_from_source only includes the data read from source during the reads
+    // It ignores the data read by BookKeeper in async warmup
+    public long getTotal_system_source_MB_read()
```
sopel39, line 137: why detailed and not basic?

stagraqubole: Basic stats are covering stats from client perspective only, i.e. counters about things that happen during the read, and are mainly powering hit-ratio. Adding system metric there can be confusing to users.

PR #446, opened 2020-08-26
Remove shading of thrift lib from submodule

suryanshagnihotri:


stagraqubole:
Is this tested in both standalone and embedded modes?

stagraqubole (approved):

PR #450, opened 2020-09-09
upgrade thrift library

suryanshagnihotri:
Upgrading thrift library to the latest one as the current one has security vulnerabilities.


stagraqubole:
This will need thorough testing, with embedded mode and stand alone mode. And also some perf testing.

PR #430, opened 2020-07-13
Add custom metrics reporter interface

yashaswaj:

PR #457, opened 2020-09-23
fix double check lock on cluster manager

suryanshagnihotri:


rubix-bookkeeper/src/main/java/com/qubole/rubix/bookkeeper/BookKeeper.java
```diff
@@ -110,7 +110,7 @@
 
   protected static Cache<String, FileMetadata> fileMetadataCache;
   private static LoadingCache<String, FileInfo> fileInfoCache;
-  protected ClusterManager clusterManager;
+  protected volatile ClusterManager clusterManager;
```
harmandeeps, line 113: can you please describe the issue?

suryanshagnihotri: I see usecase similar to [this](https://github.com/qubole/rubix/pull/430#discussion_r456235851) comment.

harmandeeps: I see, if possible can we go with early initialization as mentioned in some other links? 

Using volatile may have adverse performance implications. So, if we can avoid it may be by creating a rubix config to tell it before starting the cluster type and then we do an early initialization. This way we may get rid of double lock check pattern.

stagraqubole: We dont have a good way of initializing objects in rubix because of they way rubix integrates with engine. It gets added as a library and there is no separate initialisation phase for rubix. It is possible in embedded mode where there is a initialisation phase but not generally. 
I will merge this change, in case you have ideas about doing it the other way please open a new issue.

PR #458, opened 2020-09-23
shade com.google in build module

suryanshagnihotri:
build will fail in the client using rubix build as jar with duplicate class.
Made similar to presto-shaded module.

### DIFF
--- a/rubix-build/pom.xml
+++ b/rubix-build/pom.xml
@@ -84,6 +84,13 @@
                                     <pattern>org.apache.thrift</pattern>
                                     <shadedPattern>org.apache.thrift.shaded</shadedPattern>
                                 </relocation>
+                                <relocation>
+                                    <pattern>com.google</pattern>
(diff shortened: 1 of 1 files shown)


rubix-build/pom.xml
```diff
@@ -84,6 +84,13 @@
                                     <pattern>org.apache.thrift</pattern>
                                     <shadedPattern>org.apache.thrift.shaded</shadedPattern>
                                 </relocation>
+                                <relocation>
+                                    <pattern>com.google</pattern>
+                                    <shadedPattern>com.google.shaded</shadedPattern>
+                                    <excludes>
```
harmandeeps, line 90: any reason to exclude classes under this package?

suryanshagnihotri: It was done in this [PR](https://github.com/qubole/rubix/pull/415), so I added in build module too.

harmandeeps: got it.

harmandeeps (approved):

Issue #462, opened 2020-10-01
Avoid caching presto worker worker nodes


JamesRTaylor:
After a presto worker nodes gets deregistered from the coordinator. because it is no longer responsive, we've noticed that it still receives cache requests to be read from for a period of time. This can lead to internal errors for the query and/or slower response times. Instead, since the rubix presto cluster manager has a handle to the presto node manager, it can get the worker nodes directly from it, sorting and adjusting the node index only when the nodes have changed.

JamesRTaylor:
@stagraqubole - I can take this one if you'd like.

stagraqubole:
Sure @JamesRTaylor

Issue #454, opened 2020-09-18
Available cache space is computed before cache directories are cleaned up during startup


sopel39:
See `BookKeeper#BookKeeper(org.apache.hadoop.conf.Configuration, com.qubole.rubix.common.metrics.BookKeeperMetrics, com.google.common.base.Ticker)`.
Inside there is,
```
    initializeCache(conf, ticker);
    cleanupOldCacheFiles(conf);
```
However, `initializeCache` uses `java.io.File#getUsableSpace` to compute cache size. Therefore when node is restarted with let's say 80% cache partition occupied before, Rubix will use at most only 20% of remaining space.

Should these method calls be revered?

stagraqubole:
Yes, this needs to be fixed.

harmandeeps:
Fixed in #455. Closing this issue. 
Thanks!

sopel39:
Will you create PR to PrestoSQL once Rubix is released?

harmandeeps:
@sopel39 : yeah, I will create the PR.

PR #465, opened 2020-10-06
Avoid caching presto worker nodes

JamesRTaylor:


JamesRTaylor:
@stagraqubole - not sure about the test integration failure.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
```diff
@@ -48,170 +28,51 @@
 /**
  * Created by stagra on 14/1/16.
  */
-public class PrestoClusterManager extends ClusterManager
+public class PrestoClusterManager extends SyncClusterManager
```
stagraqubole, line 31: We need the other form too because there is a standalone mode where Presto's NodeManager is not available. We should create a new one say SyncPrestoClusterManager as a copy of what you have and keep PrestoClusterManager based on AsyncClusterManager

rubix-prestosql/src/test/java/com/qubole/rubix/prestosql/TestingNodeManager.java
```diff
@@ -0,0 +1,287 @@
...
+import io.prestosql.spi.NodeManager;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.conf.Configuration;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.lang.reflect.Type;
+import java.net.*;
```
stagraqubole, line 31: nit: expand the imports

rubix-prestosql/src/test/java/com/qubole/rubix/prestosql/TestingNodeManager.java
```diff
@@ -0,0 +1,287 @@
...
+  private static Log LOG = LogFactory.getLog(TestingNodeManager.class);
+  private static final int DEFAULT_SERVER_PORT = 8081;
+  private static final String DEFAULT_USER = "rubix";
+  public static final String SERVER_PORT_CONF_KEY = "caching.fs.presto-server-port";
+
+  private final String serverAddress;
+  private final Node currentNode;
+  private final int serverPort;
+
+  public TestingNodeManager(Configuration conf, Node currentNode)
```
stagraqubole, line 46: I guess you wont need this once you have Sync and Async implementations of PrestoClusterManager

rubix-spi/src/main/java/com/qubole/rubix/spi/SyncClusterManager.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.Set;
+
+public abstract class SyncClusterManager extends ClusterManager
+{
+  private static Log log = LogFactory.getLog(SyncClusterManager.class);
+
+  private Set<String> currentNodes;
+
+  protected abstract boolean hasStateChanged();
```
stagraqubole, line 26: Current model will spike up CPU usage a lot. The ClusterManager methods are very frequently called, e.g. `getCurrentNodeName` will be called for every MB of data being read via the worker node. This converts to 1k calls per GB to `hasStateChanged` which does comparisons internally.

I think the optimal way would be work with Presto team to add the logic in Presto's NodeManager to return info about state change e.g. it could store a version which increments every time there is a change in membership. Based on that version, we can hasStateChanged value here.

stagraqubole (changes requested):

rubix-spi/src/main/java/com/qubole/rubix/spi/SyncClusterManager.java
JamesRTaylor: I've confirmed that the hasStateChanged() method returns on an == check of the existing and new worker nodes. There's no extra overhead for this. The only time extra work is done is when the worker nodes change.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
JamesRTaylor: I created an implementation of Presto's NodeManager that can be used for standalone mode. We can move that out of the test side of things if you need it. This provides a much cleaner separation of concerns IMHO.

rubix-prestosql/src/test/java/com/qubole/rubix/prestosql/TestingNodeManager.java
JamesRTaylor: done

rubix-spi/src/main/java/com/qubole/rubix/spi/SyncClusterManager.java
stagraqubole: Wouldn't the check be on `ImmutableSet.equals` which is not a simple `==` check?

JamesRTaylor: The == check is the first line of the `ImmutableSet.equals` implementation.

stagraqubole: Ok, so you confirmed that over a period of time if cluster state does not change then Presto's NodeManager returns same object? That is a dependency on undocumented behaviour of NodeManager, we will need to way/test to catch if this behaviour changes in future otherwise slowness will creep in at runtime.

stagraqubole: Looking at https://github.com/prestosql/presto/blob/c4ef3a95309686d778c9f319d7e91be8b6a6faaa/presto-main/src/main/java/io/prestosql/metadata/DiscoveryNodeManager.java#L144, the Set object would change every 5sec

JamesRTaylor: Yes, every 5 seconds, the `ImmutableSet.equals` would do an equals check on its members and again return true (unless workers were added or removed) and further processing would be stopped. The equals check on the Node is a string compare on the node identifier, so not an expensive check.

JamesRTaylor: > Ok, so you confirmed that over a period of time if cluster state does not change then Presto's NodeManager returns same object?
Yes, I confirmed this with logging that I've since removed. Let me see if I can add a unit test that confirms too.

stagraqubole: Ok, every 5sec there will be extra work to compare all the elements of the Set. I still think that doing it in conjunction with Presto NodeManager, something like a callback that NodeManager calls when it sees change in membership would work out best.
The problem with current approach, multiple comparisons every 5sec, would amplify if we remove the synchronisation you have added for all methods of SyncClusterManager. Given that there will be 100s of parallel requests to these methods (as many as task runner threads running in Presto), current code will serialise them which will definitely have perf impact. And as soon as you remove sync, you will have the problem of these 100s of threads (potentially all but may not be always all) doing the extra comparisons every 5sec. So a solution that is deeply integrated with Presto would be better.

Tagging @sopel39 @losipiuk to get inputs if it would be acceptable to add `stateChangeListener` to `InternalNodeManager` in Presto.

JamesRTaylor: With the enormity of everything  rubix is doing, there’s no way comparing 100 strings every five seconds would add any measurable overhead. How about running it with presto to measure it?

sopel39: > I think the optimal way would be work with Presto team to add the logic in Presto's NodeManager to return info about state change

This just shifts the problem to Presto. Workers list is not a real-time list and requires some time to refresh (I think discovery requests are sent every 1s, but there is also blacklist of failed nodes refreshed independently).

On a general note, I'm not sure what problem are we trying to fix? Is that an issue that we cache workers list for 5s? Workers should join/depart cluster much less frequently. Cluster with nodes changing every 5s won't be stable anyway.

Even if the list is not current, queries should not fail.

JamesRTaylor: @sopel39 - this issue (https://github.com/qubole/rubix/issues/462) is a secondary issue in that it causes more queries to fail than otherwise would with caching enabled when a node becomes unresponsive and/or is subsequently restarted/terminated. The reason is that rubix is caching the set of worker nodes and using this to determine where to send cache read requests. Because the set is cached by rubix instead of always getting it from the NodeManager, more cache read requests will be sent to nodes that are down which leads to more query failures with an internal error reason. We've observed this behavior consistently on a heavily loaded test cluster, while without caching enabled under heavy load, queries instead fail with an insufficient resources timeout. Preventing the proliferation of these internal error type failures is important because otherwise engineers that are on call start to get paged.

This PR solves this by always getting the list of worker nodes from the NodeManager and only doing the additional computation needed by rubix when the set of worker nodes changes. @stagraqubole is worried that this will cause too much overhead and is asking if it's feasible to have some kind of callback in NodeManager when the set of active workers changes. Looking closer at DiscoveryNodeManager, though it [polls for the node state every 5 seconds](https://github.com/prestosql/presto/blob/c4ef3a95309686d778c9f319d7e91be8b6a6faaa/presto-main/src/main/java/io/prestosql/metadata/DiscoveryNodeManager.java#L144), it seems that it only updates the set of nodes [if they've changed](https://github.com/prestosql/presto/blob/c4ef3a95309686d778c9f319d7e91be8b6a6faaa/presto-main/src/main/java/io/prestosql/metadata/DiscoveryNodeManager.java#L263). Irregardless of this, even if it did update every 5 seconds, I don't think the string compare of an equals operation occurring every 5 seconds would add any overhead.

The primary issue (which is being worked on by @stagraqubole) is [fixing the disk usage accounting](https://github.com/qubole/rubix/pull/464) to prevent out of disk space issues.

Other primary issues are:
- [Preventing worker nodes from being overwhelmed](https://github.com/qubole/rubix/issues/469) by dynamically sizing the thread pools. We've worked around this by configuring the thread pools down.
- [Removing cache files](https://github.com/qubole/rubix/issues/454) prior to calculating allowed disk usage. We've worked around this by changing our presto startup script.

JamesRTaylor:
@shubhamtagra - I can put this change behind a config flag so that it maintains the current implementation if you're worried about it. We'd really like to have this capability it in the next release to test out in conjunction with your other fixes.

JamesRTaylor:
WDYT @shubhamtagra?

JamesRTaylor:
@shubhamtagra - I've updated the PR as we've discussed. Please review when you have a chance.

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
```diff
@@ -12,300 +12,51 @@
...
-  @Override
-  public List<String> getNodesInternal()
-  {
-    if (nodeManager != null) {
-      return getNodesFromNodeManager();
-    }
-
-    try {
-      URL allNodesRequest = getNodeUrl();
-      URL failedNodesRequest = getFailedNodeUrl();
```
shubhamtagra, line 82: This part needs to be retained. For the mode of rubix running as separate process NodeManager object will not be available from Presto.

shubhamtagra (changes requested):
Just one comment about keeping the existing code in PrestoClusterManager to get worker nodes list over http.

To use SyncPrestoCM, you plan to use `rubix.cluster.manager.presto.class` config?

rubix-prestosql/src/main/java/com/qubole/rubix/prestosql/PrestoClusterManager.java
shubhamtagra: Missed StandaloneNodeManager part.

shubhamtagra (approved):

JamesRTaylor:
> To use SyncPrestoCM, you plan to use rubix.cluster.manager.presto.class config?

Yes, exactly.

Thanks again for the code reviews.

shubhamtagra:
Merged, thanks @JamesRTaylor. I will release a new version so that this could be picked up in Presto.

PR #471, opened 2020-11-18
Use NodeManager to get current host name and address for prestosql

JamesRTaylor:


JamesRTaylor:
@shubhamtagra - sorry for the trouble, but would it be possible to get a new rubix release with this fix in it?
cc @harmandeeps

shubhamtagra:
@JamesRTaylor This part gets missed out in testing. It was added when you saw that right hostname was not getting picked in your setup. The code lgtm, can you please also add a simple UT which checks that expected hostname is returned by the PrestoCM. Post that, I will make a release right after merging.

JamesRTaylor:
I've added the test, @shubhamtagra. Thanks for the reviews.

shubhamtagra:
Thanks, merged.

0.3.20 has been released.

PR #473, opened 2021-02-01
Add presto dependency to presto-shaded module

agrawalreetika:
This is needed to enable Rubix caching support for presto. Right now Rubix shaded jar which is being pulled from the maven central repo doesn't include dependency from the 'rubix-presto' module.


agrawalreetika:
@shubhamtagra @abhishekdas99  Could you please review this, This is needed to enable Rubix cache for PrestoDB.

